PR #77, opened 2021-10-03
Feature/navgation body builder

### LINKED ISSUE #76: NavigationBody builder
**Is your feature request related to a problem? Please describe.**
To reduce memory consumption, we should be able to lazily build navigation body items

**Describe the solution you'd like**
Like ListView.builder, provide an index based builder to NavigationBody

**Describe alternatives you've considered**
I considered (and am using) a prebuild list of widgets instead of a builder

NicolaVerbeeck:
This adds a builder variant to NavigationBody. Fixes #76

## Pre-launch Checklist

- [X] I have run `dartfmt` on all changed files  
- [X] I have addressed all analyzer warnings as best I could
- [X] I have run `flutter pub publish --dry-run` and addressed any warnings

### DIFF
--- a/lib/src/controls/navigation/navigation_view/body.dart
+++ b/lib/src/controls/navigation/navigation_view/body.dart
@@ -24,6 +24,9 @@ class NavigationBody extends StatefulWidget {
         itemCount = null,
         super(key: key);
 
+  /// Creates a navigation body that uses a builder to supply child pages
+  ///
+  /// [index] must be greater than 0 and less than [itemCount] if it is provided
   const NavigationBody.builder({
     Key? key,
     required this.index,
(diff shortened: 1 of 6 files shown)


bdlukaa:
Shouldn't we be using `Listview.builder` when itembuilder not null? That would increase performance

NicolaVerbeeck:
I don't think so, we are using a building pattern anyway in the class by only building the widget at the current index. A listview builder does not make sense here

lib/src/controls/navigation/navigation_view/body.dart
```diff
@@ -13,18 +13,45 @@ class NavigationBody extends StatefulWidget {
   const NavigationBody({
     Key? key,
     required this.index,
-    required this.children,
+    required List<Widget> children,
     this.transitionBuilder,
     this.animationCurve,
     this.animationDuration,
   })  : assert(index >= 0 && index <= children.length),
+        // ignore: prefer_initializing_formals
```
bdlukaa, line 21: use ignore for file for both

bdlukaa (approved):

lib/src/controls/navigation/navigation_view/body.dart
NicolaVerbeeck: Done

PR #97, opened 2021-11-20
Updated border color for Expander

tobiasht:
Updated border color for Expander


tobiasht:
Addtional things that should be looked at:
1. Right border still shimmers when hovering over expanded button (when content is expanded). 
2. Change divider color between Header and Content for darkmode.
3. Add a theme preference for the expander to change background color of header and content

bdlukaa (approved):

PR #121, opened 2021-12-26
Update style for Mica and Acrylic

Kapranov98:
Updated styles to fully support Mica and Acrylic. I couldn't transfer the navViewPaneBackgroundColor variable from ThemeData to NavigationPaneThemeData :( 
Examples of changes:

### Before (Mica):
изображение
изображение

### After (Mica):
изображение
изображение

### Before: (Acrylic):
изображение
изображение

### After (Acrylic):
изображение
изображение

## Pre-launch Checklist

- [x] I have run `dartfmt` on all changed files  
- [x] I have addressed all analyzer warnings as best I could
- [x] I have run `flutter pub publish --dry-run` and addressed any warnings


example/lib/main.dart
```diff
@@ -53,6 +54,11 @@ void main() async {
       [TargetPlatform.windows, TargetPlatform.linux]
           .contains(defaultTargetPlatform)) {
     await flutter_acrylic.Window.initialize();
+    await flutter_acrylic.Window.setEffect(
+      effect: WindowEffect.mica,
+      dark: darkMode,
+    );
```
bdlukaa, line 60: This shouldn't be done here

example/lib/main.dart
```diff
@@ -85,6 +91,7 @@ class MyApp extends StatelessWidget {
           initialRoute: '/',
           routes: {'/': (_) => const MyHomePage()},
           theme: ThemeData(
+            micaBackgroundColor: Colors.transparent,
```
bdlukaa, line 94: This shouldn't be handled this way. We should check what's the current `AppTheme.acrylicEffect` to get the best color for each.

When this is transparent, we get a weird effect when it's `WindowEffect.disabled`

lib/src/controls/form/text_box.dart
```diff
@@ -686,7 +688,7 @@ class _TextBoxState extends State<TextBox>
       child: IgnorePointer(
         ignoring: !enabled,
         child: AnimatedContainer(
-          duration: theme.fasterAnimationDuration,
+          duration: const Duration(milliseconds: 0),
```
bdlukaa, line 691: Why disabling the animation?

lib/src/controls/inputs/buttons/button.dart
```diff
@@ -60,12 +60,13 @@ class Button extends BaseButton {
...
         if (states.isDisabled) return theme.disabledColor;
-        return ButtonThemeData.buttonColor(theme.brightness, states).basedOnLuminance().toAccentColor()[
+        return theme.brightness.isLight ? Colors.black : Colors.white;
+        /*return ButtonThemeData.buttonColor(theme.brightness, states).basedOnLuminance().toAccentColor()[
             states.isPressing
                 ? theme.brightness.isLight
                     ? 'lighter'
                     : 'dark'
-                : 'normal'];
+                : 'normal'];*/
```
bdlukaa, line 69: These commented lines can be removed

lib/src/controls/inputs/buttons/theme.dart
```diff
@@ -220,20 +220,20 @@ class ButtonThemeData with Diagnosticable {
...
     } else {
       if (states.isPressing) {
-        color = const Color(0xFF272727);
+        color = const Color.fromRGBO(255, 255, 255, 0.0837);
       } else if (states.isHovering) {
-        color = const Color(0xFF323232);
+        color = const Color.fromRGBO(255, 255, 255, 0.0837);
       } else {
-        color = const Color(0xFF2b2b2b);
+        color = const Color.fromRGBO(255, 255, 255, 0.0605);
```
bdlukaa, line 236: I don't feel like these colors are correct. @WinXaito what do you think?

lib/src/controls/inputs/radio_button.dart
```diff
@@ -244,8 +244,8 @@ class RadioButtonThemeData with Diagnosticable {
           color: states.isPressing
               ? backgroundColor
               : states.isHovering
-                  ? backgroundColor.withOpacity(0.8)
-                  : backgroundColor.withOpacity(0.0),
+                  ? style.inactiveColor.withOpacity(0.1)
```
bdlukaa, line 247: This doesn't look good on light theme, in my opinion

lib/src/styles/theme.dart
```diff
@@ -172,6 +172,8 @@ class ThemeData with Diagnosticable {
   final Color scaffoldBackgroundColor;
   final Color acrylicBackgroundColor;
   final Color micaBackgroundColor;
+  final Color dialogBackgroundColor;
+  final Color navViewPaneBackgroundColor;
```
bdlukaa, line 176: These should be handled on `NavigationPaneThemeData` and `DialogThemeData`

bdlukaa (changes requested):
For some reason Mica doesn't work for me. Acrylic looks good!

lib/src/controls/form/text_box.dart
Kapranov98: If the animation is on, then when you click on the field, a flashing in a different color occurs. I looked at the XAML Controls Gallery and did not notice the animation for this action there, so I disabled it.

lib/src/controls/inputs/buttons/theme.dart
Kapranov98: These colors are taken from Microsoft's Figma guidelines and have been slightly tweaked to better match WinUI. 
изображение
изображение

lib/src/controls/inputs/radio_button.dart
Kapranov98: I made the same color on hover as Checkbox and ToggleSwitch 
изображение

example/lib/main.dart
Kapranov98: Do I understand correctly that then it is better to use this option?
`micaBackgroundColor: appTheme.acrylicEffect == WindowEffect.acrylic || appTheme.acrylicEffect == WindowEffect.mica ? Colors.transparent : null,`

lib/src/controls/form/text_box.dart
WinXaito: I don’t decide if we disable animation or not (@bdlukaa take the decision), but if you remove the animation, don’t put a duration of 0 and remove totally the AnimatedContainer

PR #191, opened 2022-02-19
Example app

bdlukaa:
Fixes #190

- Added `window_manager` to manage the window
- Recreated the windows folder

## Pre-launch Checklist

- [x] I have run `dartfmt` on all changed files  
- [ ] I have updated `CHANGELOG.md` with my changes  
- [ ] I have run "optimize/organize imports" on all changed files
- [ ] I have addressed all analyzer warnings as best I could
- [ ] I have added/updated relevant documentation
- [ ] I have run `flutter pub publish --dry-run` and addressed any warnings

PR #192, opened 2022-02-19
Fix linux build

WinXaito:
With this fix, the app can be build.

image

But there is some bug on the example app for linux:

- Windows cross are shown
- The app is transparent some seconds after start (not tested in release, only in debug)

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -32,8 +32,6 @@ bool get isDesktop {
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  SystemTheme.accentInstance;
-
   setPathUrlStrategy();
 
   if (isDesktop) {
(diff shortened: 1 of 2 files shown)


example/lib/main.dart
```diff
@@ -32,8 +32,6 @@ bool get isDesktop {
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  SystemTheme.accentInstance;
```
WinXaito, line 35: I'm not sure about that, this had an utility ???

bdlukaa: Yes! It's used to get the current accent color of the device.

bdlukaa:
Closing in favor of https://github.com/bdlukaa/fluent_ui/pull/191

PR #209, opened 2022-03-03
Fix disabled color foreground on FilledButton

WinXaito:
## Pre-launch Checklist

- [ ] I have run `dartfmt` on all changed files  
- [ ] I have addressed all analyzer warnings as best I could
- [ ] I have run `flutter pub publish --dry-run` and addressed any warnings

According to the normal button : https://github.com/bdlukaa/fluent_ui/blob/master/lib/src/controls/inputs/buttons/button.dart#L62

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 Date format: DD/MM/YYYY
 
+## [3.9.2] [dd/mm/yyyy]
+
+- Fix: [#207](https://github.com/bdlukaa/fluent_ui/pull/207) FilledButton disabled foreground
+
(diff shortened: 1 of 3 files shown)


bdlukaa:
I believe the color shouldn't be the same as `Button`

For light mode:
light_toggle_button
And for dark mode:
dark_toggle_button

WinXaito:
Currently I just fixed the foreground color.
So we need to change the background color.

but I agree with you.

btw how is on Microsoft  fluent ?

bdlukaa:
I'm no longer getting based on the web fluent documentation. I take examples on the Windows native apps I find

bdlukaa (approved):

PR #214, opened 2022-03-05
Replace WindowButtons & Add Confirm before closing

lijy91:
## Pre-launch Checklist

- [x] I have run `dartfmt` on all changed files  
- [x] I have addressed all analyzer warnings as best I could
- [x] I have run `flutter pub publish --dry-run` and addressed any warnings


WinXaito:
I think the prevent on close is not really useful (not a feature from fluent), and I don’t like when a software does nothing and want a confirmation for close.

bdlukaa:
> I think the prevent on close is not really useful (not a feature from fluent), and I don’t like when a software does nothing and want a confirmation for close.

I wanted to add this a long time ago. I believe it's a good way to showcase the `ContentDialog` as well.

bdlukaa (approved):
LGTM!

Issue #224, opened 2022-03-13
Vertical scrolling and selection changed callback for TreeView


klondikedragon:
I'm using TreeView with potentially 10,000 or 100,000 items in it, so it needs to be able to scroll vertically.

The TreeView should optionally be able to vertically scroll its content instead of just growing to meet the size of all items.

It should also be more performant for very large lists so that every item isn't rendered every time, but only those items in the current viewport. Ideally something like ListView would be used as it is optimized to do this especially if items are a fixed vertical size.

Also, my items are coming from a database query where each row in the response becomes an item in the TreeView. The widget content of a TreeViewItem may be the same between different items even though they come from different database rows, and I need to be able to distinguish between different items (e.g., by associating the primary key of a given row with a given TreeViewItem). I also need to be able to be notified when the selection changes. `onItemInvoked` is not always invoked with the selection state changes in a multi-select TreeView.

I'll submit a PR that is a suggested implementation of this feature request.

klondikedragon:
Implemented by #225

PR #225, opened 2022-03-13
Vertical scrolling for TreeView and other enhancements

klondikedragon:
* `TreeView` is now implemented using `ListView.builder`, allowing for the possibility of vertical scrolling and better efficiency for very large lists of items
* `shrinkWrap` defaults to true, so the old behavior is still the default
* `TreeViewItem.value` field added to optionally associate any value (e.g., a primary key value) with each item
* `TreeView.onSelectionChanged` callback added to be notified under all circumstances when the selection changes, and to be given the full list of the item(s) that are now currently selected (especially efficient for single selection `TreeView` when the item list is very long).
* Fixes vertical alignment issue if you have some children that are expandable and others that are not.

Implements #224.

Successfully tested with a tree with around 10,000 items with more than 10 levels of nesting, fully expanded. Scrolling/rendering performance is good in this case when setting `cacheExtent: 50`, `usePrototypeItem: true`, and `shrinkWrap: false`.

Here is a visual on the large list with scrolling with the alignment bug before the fix:
image

And here is the same list with the alignment bug fixed:
image

## Pre-launch Checklist

- [X] I have run `dartfmt` on all changed files  
- [X] I have addressed all analyzer warnings as best I could
- [X] I have run `flutter pub publish --dry-run` and addressed any warnings


lib/src/controls/navigation/tree_view.dart
```diff
@@ -309,12 +365,39 @@ class TreeView extends StatefulWidget {
...
   /// [kTreeViewLoadingIndicator] is used by default
   final Widget loadingWidget;
 
+  /// Same meaning as [ListView.shrinkWrap]
+  final bool shrinkWrap;
+
+  /// Same meaning as [ListView.cacheExtent]
+  final double? cacheExtent;
+
+  /// Same meaning as [ListView.itemExtent]
```
bdlukaa, line 388:
```suggestion
  /// {@macro flutter.widgets.list_view.itemExtent}
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -309,12 +365,39 @@ class TreeView extends StatefulWidget {
...
+  final double? itemExtent;
+
+  /// Same meaning as [ListView.addRepaintBoundaries]
```
bdlukaa, line 391:
```suggestion
  /// Whether to wrap each child in a [RepaintBoundary].
  ///
  /// Typically, children in a scrolling container are wrapped in repaint
  /// boundaries so that they do not need to be repainted as the list scrolls.
  /// If the children are easy to repaint (e.g., solid color blocks or a short
  /// snippet of text), it might be more efficient to not add a repaint boundary
  /// and simply repaint the children during scrolling.
  ///
  /// Defaults to true.
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -533,7 +645,7 @@ class _TreeViewItem extends StatelessWidget {
                       ),
                     ),
                   Padding(
-                    padding: const EdgeInsetsDirectional.only(start: 18.0),
+                    padding: const EdgeInsetsDirectional.only(start: 10.0),
```
bdlukaa, line 648: why was this value changed?

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Date format: DD/MM/YYYY
 
 - Improves `icons.dart` formatting and its generation.
 - Fix: [#207](https://github.com/bdlukaa/fluent_ui/pull/207) FilledButton disabled foreground
+- TreeView now implemented with ListView and optionally allows vertical scrolling
```
bdlukaa, line 7:
```suggestion
- TreeView now implemented with ListView and optionally allows vertical scrolling ([#255](https://github.com/bdlukaa/fluent_ui/pull/225))
```

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Date format: DD/MM/YYYY
...
+- TreeViewItem supports custom value and TreeView supports onSelectionChanged callback
```
bdlukaa, line 8:
```suggestion
- TreeViewItem now has custom primary key
- Added `onSelectionChanged` callback to `TreeView`
```

bdlukaa (changes requested):

example/lib/screens/others.dart
```diff
@@ -327,24 +372,30 @@ class _OthersState extends State<Others> {
...
                   maxWidth: 350,
                 ),
                 decoration: BoxDecoration(
                   border:
                       Border.all(color: FluentTheme.of(context).inactiveColor),
                 ),
-                child: TreeView(items: items),
+                child: TreeView(
+                  // make a deep copy of the item list, to keep UI state separate
+                  items: items.map((i) => TreeViewItem.from(i)).toList(),
```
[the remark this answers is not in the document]
klondikedragon: Good catch. You're right it shouldn't copy the list in the widget build -- it wouldn't preserve state in all cases, and would apply lots of pressure on the GC. I'll move this to `initState` instead where `lazy` is currently initialized.

lib/src/controls/navigation/tree_view.dart
klondikedragon: Good point. Before I had added the `_anyExpandableSiblings` concept it was always adding an extra 8 pixels of horizontal padding for every item, even for siblings who were all not expandable, so this change was to offset that in the other direction. Now that the `_anyExpandableSiblings` value is calculated, it only needs to add the extra padding if there is indeed one sibling who is expandable within that given part of the tree. So this can be changed back to what it was.

klondikedragon:
@bdlukaa - changes made, see what you think. Thanks.

lib/src/controls/navigation/tree_view.dart
```diff
@@ -396,6 +511,11 @@ class _TreeViewState extends State<TreeView> {
                         ..executeForAllParents((p) => p?.updateSelected());
                     }
                   });
+                  if (onSelectionChanged != null) {
+                    var selectedItems =
```
bdlukaa, line 515:
```suggestion
                    final selectedItems =
```

bdlukaa (changes requested):
`var` -> `final`

lib/src/controls/navigation/tree_view.dart
```diff
@@ -309,6 +309,14 @@ extension TreeViewItemCollection on List<TreeViewItem> {
       child.children.executeForAll(callback);
     }
   }
+
+  Iterable<TreeViewItem> whereForAll(bool Function(TreeViewItem element) t) {
```
bdlukaa, line 313: could be private

klondikedragon: my preference would be to keep it public. similar to how `executeForAll` is public, it's handy to have `whereForAll` be public as well. It's useful for my application code to easily filter the list of all items in the tree for some condition (e.g., doing a search over items).

bdlukaa (approved):
LGTM

Issue #231, opened 2022-03-19
Implement command bars


klondikedragon:
Implement command bars as described [in the spec](https://docs.microsoft.com/en-us/windows/apps/design/controls/command-bar). For example:

![image](https://docs.microsoft.com/en-us/windows/apps/design/controls/images/controls-appbar-icons.png)

Initially it may seem that a Command Bar is just a row of buttons; however, there are certain expectations on height and spacing. More importantly, the logic can get complex once you consider the open/closed state capability as well as automatic dynamic overflow of items into the "secondary commands" menu (shown when "..." button is clicked).

Command bars will be used heavily in my project and so I've implemented a basic version as a starting point. I'll share a PR to see if it is an OK starting point. It does not have dynamic overflow, but instead provides either clipping, horizontal scrolling, or wrapping to handle overflow.

Also, it's not mentioned in the spec, but command bars are sometimes contained within a raised card. For example, the application-level command bar for desktop Microsoft Office apps looks like:

![image](https://user-images.githubusercontent.com/88254524/159104956-8f080c20-ba7e-4be8-a64a-fbf09ef466b8.png)

A `CommandBarCard` will be provided to standardize the look/feel for any "raised" command bars as well.

WinXaito:
Closed with #232

PR #232, opened 2022-03-19
Implement basic command bars

klondikedragon:
This is a first implementation of command bars as described in #231.

A `CommandBar` control provides quick access to common tasks. This could be application-level or page-level commands.

Refer to [the spec](https://docs.microsoft.com/en-us/windows/apps/design/controls/command-bar).

The current implementation does not implement the open/closed state, nor does it implement the secondary commands "more" button with dynamic overflow of items into the secondary commands menu.

Also, during implementation of the example, I realized that I needed control over the `primary` and `controller` properties of the `ListView` inside of the `TreeView`, so this also adds these properties to `TreeView`.

Here is what the examples look like with the various overflow behaviors:

image

And here is what it looks like alongside the large `TreeView`:

image

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Date format: DD/MM/YYYY
 - TreeView now implemented with ListView and optionally allows vertical scrolling ([#255](https://github.com/bdlukaa/fluent_ui/pull/225))
 - TreeViewItem now has custom primary key (`value` field)
 - Added `onSelectionChanged` callback to `TreeView`
+- Implement `CommandBar` (dynamic overflow not yet supported)
+- Add `HorizontalScrollView` helper widget, with mouse wheel horizontal scrolling
 
 ## [3.9.1] - Input Update - [25/02/2022]
 
--- a/lib/src/controls/surfaces/commandbar.dart
+++ b/lib/src/controls/surfaces/commandbar.dart
@@ -1,0 +1,178 @@
+import 'package:fluent_ui/fluent_ui.dart';
+
+/// A card with appropriate margins, padding, and elevation for it to
+/// contain one or more [CommandBar]s.
+class CommandBarCard extends StatelessWidget {
+  final Widget child;
+  final double elevation;
+  final EdgeInsetsGeometry? margin;
+  final EdgeInsets padding;
+
+  const CommandBarCard({
+    Key? key,
+    required this.child,
+    this.margin,
+    this.padding = const EdgeInsets.symmetric(horizontal: 6.0, vertical: 4.0),
+    this.elevation = 2.0,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      margin: margin,
+      child: Card(
+        padding: padding,
+        elevation: elevation,
+        child: child,
+      ),
+    );
+  }
+}
+
+/// The type of wrapping used for the items inside of the CommandBar.
+///
+///   * [CommandBarOverflowBehavior.scrolling] will cause items to scroll horizontally.
+///   * [CommandBarOverflowBehavior.noWrap] will expand the size of the CommandBar based
+///      on the size of the contained items.
+///   * [CommandBarOverflowBehavior.wrap] will wrap items onto additional lines as needed.
+///   * [CommandBarOverflowBehavior.clip] will keep items on one line and clip as needed.
+enum CommandBarOverflowBehavior {
+  scrolling,
+  noWrap,
+  wrap,
+  clip,
+  // TODO: Implement support for an overflow button and dynamically overflowing items into the "SecondaryCommands" flyout
+}
+
+/// Command bars provide quick access to common tasks. This could be
+/// application-level or page-level commands.
+///
+/// A command bar is composed of a series of [CommandBarItem]s, which each could
+/// be a [CommandBarButton] or a custom [CommandBarItem].
+///
+/// If there is not enough horizontal space to display all items, the wrapping
+/// behavior is determined by [wrapType].
+///
+/// ![CommandBar example](https://docs.microsoft.com/en-us/windows/apps/design/controls/images/controls-appbar-icons.png)
+///
+/// See also:
+///
+///   * <https://docs.microsoft.com/en-us/windows/apps/design/controls/command-bar>
+class CommandBar extends StatelessWidget {
+  final List<Widget> children;
+  final CommandBarOverflowBehavior wrapType;
+  final bool _isExpanded;
+  final ScrollController? parentVerticalScrollController;
+
+  const CommandBar(
+      {Key? key,
+      required this.children,
+      this.wrapType = CommandBarOverflowBehavior.scrolling,
+      this.parentVerticalScrollController})
+      : _isExpanded = !(wrapType == CommandBarOverflowBehavior.noWrap),
+        super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    late Widget w;
+    switch (wrapType) {
+      case CommandBarOverflowBehavior.scrolling:
+        w = HorizontalScrollView(
+          child: Row(children: children),
+          parentVerticalScrollController: parentVerticalScrollController,
+        );
+        break;
+      case CommandBarOverflowBehavior.noWrap:
+        w = Row(children: children);
+        break;
+      case CommandBarOverflowBehavior.wrap:
+        w = Wrap(
+          children: children,
+        );
+        break;
+      case CommandBarOverflowBehavior.clip:
+        w = SingleChildScrollView(
+          scrollDirection: Axis.horizontal,
+          physics: const NeverScrollableScrollPhysics(),
+          child: Row(children: children),
+        );
+        break;
+    }
+    if (_isExpanded) {
+      w = Row(children: [Expanded(child: w)]);
+    }
+    return w;
+  }
+}
+
+/// An individual control displayed within a [CommandBar]. This widget ensures
+/// that the child widget has the proper margin so the item has the proper
+/// minimum height and width expected of a control within a [CommandBar].
+class CommandBarItem extends StatelessWidget {
+  final Widget child;
+
+  const CommandBarItem({Key? key, required this.child}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      margin: const EdgeInsets.symmetric(vertical: 6.0, horizontal: 3.0),
+      child: child,
+    );
+  }
+}
+
+/// Buttons are the most common control to put within a [CommandBar].
+/// They are composed of an (optional) icon and an optional
+class CommandBarButton extends StatelessWidget {
+  final IconData? icon;
+  final Color? iconColor;
+  final String? iconSemanticLabel;
+  final TextDirection? iconTextDirection;
+  final Widget? label;
+  final VoidCallback? onPressed;
+  final VoidCallback? onLongPress;
+  final FocusNode? focusNode;
+  final bool autofocus;
+
+  const CommandBarButton(
+      {Key? key,
+      this.icon,
+      this.iconColor,
+      this.iconSemanticLabel,
+      this.iconTextDirection,
+      required this.onPressed,
+      this.onLongPress,
+      this.focusNode,
+      this.autofocus = false,
+      this.label})
+      : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return IconButton(
+      onPressed: onPressed,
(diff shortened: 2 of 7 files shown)


henry2man:
I am very interested in this component. I also have an implementation, different but perhaps it can contribute things.For example, in my version I distinguish two groups of buttons, one of which is automatically hidden when there is no horizontal space.If you want we can mark it as a draft and spend a few days reviewing it together.

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+      margin: const EdgeInsets.symmetric(vertical: 6.0, horizontal: 3.0),
+      child: child,
+    );
+  }
+}
+
+/// Buttons are the most common control to put within a [CommandBar].
+/// They are composed of an (optional) icon and an optional
+class CommandBarButton extends StatelessWidget {
+  final IconData? icon;
```
bdlukaa, line 128: Should be a `Widget`

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+    required this.child,
+    this.margin,
+    this.padding = const EdgeInsets.symmetric(horizontal: 6.0, vertical: 4.0),
+    this.elevation = 2.0,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      margin: margin,
```
bdlukaa, line 22:
```suggestion
    return Padding(
      padding: margin,
```

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+  }
+}
+
+/// The type of wrapping used for the items inside of the CommandBar.
+///
+///   * [CommandBarOverflowBehavior.scrolling] will cause items to scroll horizontally.
+///   * [CommandBarOverflowBehavior.noWrap] will expand the size of the CommandBar based
+///      on the size of the contained items.
+///   * [CommandBarOverflowBehavior.wrap] will wrap items onto additional lines as needed.
+///   * [CommandBarOverflowBehavior.clip] will keep items on one line and clip as needed.
```
bdlukaa, line 38: This can be added to the items themselves

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+  final CommandBarOverflowBehavior wrapType;
+  final bool _isExpanded;
+  final ScrollController? parentVerticalScrollController;
+
+  const CommandBar(
+      {Key? key,
+      required this.children,
+      this.wrapType = CommandBarOverflowBehavior.scrolling,
+      this.parentVerticalScrollController})
+      : _isExpanded = !(wrapType == CommandBarOverflowBehavior.noWrap),
```
bdlukaa, line 72:
```suggestion
      : _isExpanded = wrapType != CommandBarOverflowBehavior.noWrap,
```

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+  final List<Widget> children;
+  final CommandBarOverflowBehavior wrapType;
+  final bool _isExpanded;
+  final ScrollController? parentVerticalScrollController;
+
+  const CommandBar(
+      {Key? key,
+      required this.children,
+      this.wrapType = CommandBarOverflowBehavior.scrolling,
+      this.parentVerticalScrollController})
```
bdlukaa, line 71: trailing comma

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,178 @@
...
+      {Key? key,
+      this.icon,
+      this.iconColor,
+      this.iconSemanticLabel,
+      this.iconTextDirection,
+      required this.onPressed,
+      this.onLongPress,
+      this.focusNode,
+      this.autofocus = false,
+      this.label})
```
bdlukaa, line 148: trailing comma

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Date format: DD/MM/YYYY
 - TreeView now implemented with ListView and optionally allows vertical scrolling ([#255](https://github.com/bdlukaa/fluent_ui/pull/225))
 - TreeViewItem now has custom primary key (`value` field)
 - Added `onSelectionChanged` callback to `TreeView`
+- Implement `CommandBar` (dynamic overflow not yet supported)
```
bdlukaa, line 12:
```suggestion
- Implement basic `CommandBar` ([#232](https://github.com/bdlukaa/fluent_ui/pull/232))
```

lib/src/utils/horizontal_scroll_view.dart
```diff
@@ -0,0 +1,80 @@
...
+                        .parentVerticalScrollController!.positions.isNotEmpty) {
+                  widget.parentVerticalScrollController!.jumpTo(widget
+                          .parentVerticalScrollController!
+                          .positions
+                          .last
+                          .pixels -
+                      event.scrollDelta.dy);
+                }
+              }
+            }
```
bdlukaa, line 70: What does this do?

bdlukaa (changes requested):
Added a basic review. Documentation and formatting needs to be improved

lib/src/utils/horizontal_scroll_view.dart
klondikedragon: It was an ugly hack that I am not proud of! I've figured out the proper way to do this using `PointerSignalResolver` and was able to delete all references to `parentVerticalScrollController`. It now captures the mouse wheel scrolling and prevents propagation of it without having to specify any surrounding scroll controllers.

klondikedragon:
@bdlukaa - addressed comments and added `CommandBarSeparator` (corollary to `PaneItemSeparator`).

@henry2man - all help would be welcome! this could be added as an enhancement to this (whether in this PR or a follow-on) by:
* Adding a new type of `CommandBarOverflowBehavior` (something like `dynamicOverflow`)
* Adding a new widget `CommandBar` widget field `secondaryItems` (defaults to empty)
* In build if `secondaryItems` is not empty to automatically add a button using the `FluentIcons.more` icon that when clicked will present a flyout with the `secondaryItems`
* Adding logic to somehow calculate during build or layout of `CommandBar` (if overflow mode is `dynamicOverflow`) which of the normal items should instead be considered part of `secondaryItems`

Do you have any code to reference that demonstrates how to figure out which of the items can fit vs which ones should be hidden into a secondary menu?

henry2man:
> Do you have any code to reference that demonstrates how to figure out which of the items can fit vs which ones should be hidden into a secondary menu?

I have a very simple code: a width parameter in px that is evaluated during the execution of the "build" method. A  [LayoutBuilder](https://api.flutter.dev/flutter/widgets/LayoutBuilder-class.html) could be used: 

"if width is less than x then hide buttons and show menú button"

klondikedragon:
@henry2man @bdlukaa -- After some research, I figured out how to implement a fully featured command bar, including:
* Support for separate primary vs secondary command bar items, with secondary items shown in a flyout menu behind a "more" button
* Dynamic overflow behavior to put overflowing primary command bar items into the secondary area (flyout menu)
* A "compact" mode for primary command bar items, which is automatically selected if the width of the command bar is less than a specified breakpoint (off by default).
* Support for right-aligned command bars (the fluent UI spec seems to prefer this in some contexts).

This was implemented with a new `DynamicOverflow` layout widget that supports both horizontal and vertical layouts. The code for this was carefully adapted from the standard `Wrap` widget. This widget could potentially be useful in the future to implement a "more" button for horizontal or vertical navbar pane items (instead of scrolling).

In the example app, with additional examples for command bars added, it was getting cluttered on the others pane, so I added another navbar pane. I could move it back to others if you want, it was just a little awkward testing the wrapping behavior with a lot of stuff above it.

All the new functionality is best illustrated with a video:
fluent_ui_commandbar_dynamic_overflow

bdlukaa:
SO COOL! This will open possibilities for the same thing on NavigationView! I'll review ASAP

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,552 @@
...
+
+  const CommandBar({
+    Key? key,
+    required this.primaryItems,
+    this.secondaryItems = const [],
+    this.overflowItemBuilder,
+    this.overflowBehavior = CommandBarOverflowBehavior.dynamicOverflow,
+    this.compactBreakpointWidth,
+    this.mainAxisAlignment = MainAxisAlignment.start,
+    this.crossAxisAlignment = CrossAxisAlignment.center,
```
bdlukaa, line 126: on #181, Right to Left Support was added. The current `Directionality` should be taken into account

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,552 @@
...
+      case CrossAxisAlignment.end:
+        return WrapCrossAlignment.end;
+      case CrossAxisAlignment.center:
+        return WrapCrossAlignment.center;
+      case CrossAxisAlignment.stretch:
+        throw UnsupportedError(
+            "CommandBar does not support CrossAxisAlignment.stretch");
+      case CrossAxisAlignment.baseline:
+        throw UnsupportedError(
+            "CommandBar does not support CrossAxisAlignment.baseline");
```
bdlukaa, line 176:
```suggestion
            'CommandBar does not support CrossAxisAlignment.baseline');
```

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,552 @@
...
+  });
+
+  /// Builds the final widget for this display mode for this item.
+  /// This applies the widgetBuilder callback to the internally built
+  /// widget before returning.
+  Widget build(BuildContext context, CommandBarItemDisplayMode displayMode) {
+    var w = buildWidget(context, displayMode);
+    if (widgetBuilder != null) {
+      w = widgetBuilder!(context, displayMode, w);
+    }
```
bdlukaa, line 382: Currently, `buildWidget` will always be called. This could cause performance issues and memory leak. With the suggestion, `buildWidget` will only be called if necessary
```suggestion
    if (widgetBuilder != null) {
      return widgetBuilder!(context, displayMode, w);
    } else {
      return buildWidget(context, displayMode);
    }
```

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -0,0 +1,552 @@
...
+
+/// An individual control displayed within a [CommandBar]. This widget ensures
+/// that the child widget has the proper margin so the item has the proper
+/// minimum height and width expected of a control within a [CommandBar].
+abstract class CommandBarItem with Diagnosticable {
+  final Key? key;
+
+  /// Specify to allow customization of the built widget for a given
+  /// display mode. For example, to wrap a widget in a [Tooltip].
+  final CommandBarItemWidgetBuilder? widgetBuilder;
```
bdlukaa, line 369: I don't like this. The dev can extend a `CommandBarItem` and do it on the `buildWidget` method.

Something like this can be seen on https://github.com/bdlukaa/fluent_ui/blob/master/example/lib/main.dart#L300-L339

bdlukaa (changes requested):
Awesome job!

lib/src/controls/surfaces/commandbar.dart
klondikedragon: Tight-to-left Directionality is already considered in the widgets used inside the `CommandBar` (whether it's `Row` or `DynamicOverflow`). I just needed to change the text "right-aligned" to "end-aligned" in the example so it's not confusing when right-to-left mode is enabled:
image

For the secondary flyout menu content, I did notice that `ListTile` needed its padding fixed in RTL mode, so I fixed that.

lib/src/controls/surfaces/commandbar.dart
klondikedragon: Now that this logic has been moved into a separate type of item, `CommandBarBuilderItem`, this logic will only be used when this type of item is explicitly used, which requires a non-null callback, so it should solve for this as well.

bdlukaa (approved):
Awesome work! LGTM

WinXaito:
Incredible !

bdlukaa:
@klondikedragon Hello! I'm having issues with https://github.com/bdlukaa/fluent_ui/pull/277. It uses `DynamicOverflow`, but I need to ensure the current selected item is being displayed. Can you help me on that? If you want, we can talk on discord :)

PR #235, opened 2022-03-22
Add isCompact field to CommandBar

klondikedragon:
Small follow-on tweak to #232 to add an explicit field `isCompact` that can indicate that the primary display area should display the items in compact mode, without relying on `compactBreakpointWidth`. While `compactBreakpointWidth` is convenient, it was breaking things in some of my more complex layouts that already had a `LayoutBuilder` in a parent widget higher in the org tree. This allows use of the `compactBreakpointWidth` for convenience in most cases, or complete control over compact mode using `isCompact` otherwise. Default behavior has not changed.

### DIFF
--- a/lib/src/controls/surfaces/commandbar.dart
+++ b/lib/src/controls/surfaces/commandbar.dart
@@ -98,6 +98,20 @@ class CommandBar extends StatefulWidget {
   /// using [CommandBarItemDisplayMode.inPrimary].
   final double? compactBreakpointWidth;
 
+  /// If [compactBreakpointWidth] is null, then specifies whether or not
+  /// primary items should be displayed in compact mode
+  /// ([CommandBarItemDisplayMode.inPrimaryCompact]) or normal mode
+  /// [CommandBarItemDisplayMode.inPrimary].
+  ///
+  /// This can be useful if the CommandBar is used in a setting where
+  /// [compactBreakpointWidth] cannot be used (i.e. because using
+  /// [LayoutBuilder] cannot be used in a context where the intrinsic
+  /// height is also calculated), and you want to specify whether or not
+  /// the primary items should be compact or not.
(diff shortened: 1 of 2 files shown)


lib/src/controls/surfaces/commandbar.dart
```diff
@@ -309,7 +324,10 @@ class _CommandBarState extends State<CommandBar> {
   @override
   Widget build(BuildContext context) {
     if (widget.compactBreakpointWidth == null) {
-      return _buildForPrimaryMode(context, CommandBarItemDisplayMode.inPrimary);
+      var displayMode = (widget.isCompact ?? false)
```
bdlukaa, line 327:
```suggestion
      final displayMode = (widget.isCompact ?? false)
```

lib/src/controls/surfaces/commandbar.dart
```diff
@@ -98,6 +98,20 @@ class CommandBar extends StatefulWidget {
...
+  /// ([CommandBarItemDisplayMode.inPrimaryCompact]) or normal mode
+  /// [CommandBarItemDisplayMode.inPrimary].
+  ///
+  /// This can be useful if the CommandBar is used in a setting where
+  /// [compactBreakpointWidth] cannot be used (i.e. because using
+  /// [LayoutBuilder] cannot be used in a context where the intrinsic
+  /// height is also calculated), and you want to specify whether or not
+  /// the primary items should be compact or not.
+  ///
+  /// If [compactBreakpointWidth] is not null this field is ignored.
```
bdlukaa, line 112:
```suggestion
  /// If [compactBreakpointWidth] is not `null`, this field is ignored.
```

bdlukaa (changes requested):

bdlukaa (approved):

Issue #220, opened 2022-03-09
[feature request] `Expander` without content


h3x4d3c1m4l:
**Is your feature request related to a problem? Please describe.**
As a developer creating a settings screen in my app, I would like to make it look and feel like the Windows 11 settings. For this we can use Extender, but that requires a content widget argument, while Windows 11 seem to use Extender like widgets without any content (e.g. with only a dropdown box at the end).

**Describe alternatives you've considered**
Alternative layout (simple listview, or perhaps based on some type of grid), or create a subclass of Extender to implement this.

**Additional context**
![Example from Windows 11 Settings](https://user-images.githubusercontent.com/2611894/157348828-b21c7f42-956e-4535-82e4-4c33081f0192.png)

WinXaito:
You can do something like that:

```dart
Card(
  child: ListTile(
    leading: const Icon(FluentIcons.brightness),
    title: const Text('Title'),
    subtitle: const Text('Subtitle'),
    trailing: DropDownButton(title: const Text('DropDown'), items: [
      DropDownButtonItem(onTap: () {}, title: const Text('test'))
    ]),
  ),
),
```

Render:

![image](https://user-images.githubusercontent.com/8223773/157394193-ee4b4c83-c99c-453c-a4f7-7523f6cc2b41.png)

bdlukaa:
This field on the Windows Settings App is a custom component of the app, not part of the fluent design system. You can check the complete list of fluent components [here](https://github.com/microsoft/microsoft-ui-xaml/tree/main/dev).

Closing as won't fix

PR #239, opened 2022-03-24
Optional content for Expander

tobiasht:
Made the `content` parameter optional for `Expander`. An  `Expander` with no content will not be able to be expanded, but will still have the same theming as the a regular  `Expander`

## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/surfaces/expander.dart
+++ b/lib/src/controls/surfaces/expander.dart
@@ -66,7 +66,7 @@ class Expander extends StatefulWidget {
   /// of a parent Expander as shown here.
   ///
   /// ![Expander Nested Content](https://docs.microsoft.com/en-us/windows/apps/design/controls/images/expander-nested.png)
-  final Widget content;
+  final Widget? content;
 
   /// The icon of the toggle button.
   final Widget? icon;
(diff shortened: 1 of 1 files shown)


lib/src/controls/surfaces/expander.dart
```diff
@@ -198,6 +198,52 @@ class ExpanderState extends State<Expander>
   Widget build(BuildContext context) {
     assert(debugCheckHasFluentTheme(context));
     theme = FluentTheme.of(context);
+    if (widget.content == null) {
```
bdlukaa, line 201: Would this be some boilerplate code? Isn't it better to check if content is null on `SizeTransition`?

tobiasht: You might be right, but I haven't used `SizeTransition` before. Would the use of `SizeTransition` still hide the handle and prevent any actions on the object? (Hover color etc)

h3x4d3c1m4l:
I think this would fix #220.

bdlukaa:
I don't like the idea of this. It'd be better to re-style `ListTile` to match the new Win UI 3 instead. Closing

PR #240, opened 2022-03-24
Clear button autosuggest

### LINKED ISSUE #238: 🐛 AutoSuggestBox clear button does not work
**Describe the bug**
The `AutoSuggestBox`  has the ability to display a clear button allowing for setting the `TextController` to empty and removing focus. When clicking the clear button the `onPressed` callback is never called.

**To Reproduce**
1. Go to https://bdlukaa.github.io/fluent_ui/
2. Click on the auto suggest box on the left side of the screen
3. Select one of the suggested choice
4. Click on the box to get focus
5. Click on the newly appeared cleat button on the right of the field.
6. Nothing happens.

**Expected behavior**
After step 5, the field is supposed to be cleared and the focus removed form it..

**Additional context**
I copied the code of the `IconButton` used to build the clear button. It works well on it's own. So the problem is localized inside the `AutoSuggestBox` widget itself.

bdlukaa:
Fixes #238 
Updates `TextBox` colors

## Pre-launch Checklist

### DIFF
--- a/example/lib/screens/forms.dart
+++ b/example/lib/screens/forms.dart
@@ -193,15 +193,12 @@ class _FormsState extends State<Forms> {
         const SizedBox(height: 20),
         InfoLabel(
           label: 'Selectable Text',
-          child: Mica(
-            child: Padding(
-              padding: const EdgeInsets.all(8.0),
-              child: SelectableText(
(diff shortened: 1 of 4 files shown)


henry2man:
https://github.com/bdlukaa/fluent_ui/issues/238#issuecomment-1078703536
> @bdlukaa Although I have also had this issue I have seen another that may be related. When an element is chosen from the drop-down list, a change of type `TextChangedReason.suggestionChosen` is NOT being generated; instead `TextChangedReason.userInput` is always sent

shunsen:
there ia a problem described in the bug #238 and it is still exists now; 
described:When clicking the clear button the onPressed callback is never called.
expected:after click on the newly appeared clear button on the right of the field, the callback function onpressed or onChanged should be called;

PR #241, opened 2022-03-24
Restyle CommandBar Flyout

bdlukaa:
- Added `FlyoutListTile`, that uses the same base as `DropDownButtonItem.build` and the tile for text selection controls. In the future, they should be unified
- Update the `FlyoutContent`, without `Acrylic` and with the recent `FluentTheme.menuColor`

example 24_03_2022 18_01_28

@klondikedragon

## Pre-launch Checklist

- [ ] I have updated `CHANGELOG.md` with my changes   _Not necessary since this is a continuation of the CommandBars PR, that hasn't been released_

### DIFF
--- a/lib/src/controls/surfaces/flyout/flyout.dart
+++ b/lib/src/controls/surfaces/flyout/flyout.dart
@@ -163,11 +163,11 @@ class FlyoutListTile extends StatelessWidget {
             color: ButtonThemeData.uncheckedInputColor(theme, states),
             borderRadius: radius,
           ),
-          padding: const EdgeInsets.only(
+          padding: const EdgeInsetsDirectional.only(
             top: 4.0,
             bottom: 4.0,
-            left: 10.0,
-            right: 8.0,
+            start: 10.0,
+            end: 8.0,
           ),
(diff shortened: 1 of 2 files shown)


lib/src/controls/surfaces/flyout/flyout.dart
```diff
@@ -122,3 +122,106 @@ class FlyoutContent extends StatelessWidget {
...
+      builder: (context, states) {
+        final theme = FluentTheme.of(context);
+        final radius = BorderRadius.circular(4.0);
+
+        Widget content = Container(
+          decoration: BoxDecoration(
+            color: ButtonThemeData.uncheckedInputColor(theme, states),
+            borderRadius: radius,
+          ),
+          padding: const EdgeInsets.only(
```
klondikedragon, line 166: Should this potentially be `EdgeInsetsDirectional`?

klondikedragon (approved):
LGTM! one minor point to consider

lib/src/controls/surfaces/flyout/flyout.dart
bdlukaa: Nice catch!

PR #270, opened 2022-04-05
Fix overflow behavior for TreeViewItem

### LINKED ISSUE #269: Long Text widget in TreeView item overflows🐛
**Describe the bug**
If you have a `TreeViewItem` that contains content with a `Text` widget with text that is longer than the parent `TreeView` widget, it will overflow.

**To Reproduce**
Here is example modified with a tree view item with long text:
image

**Expected behavior**
Item content widgets should have a size constraint set by parent so that they properly crop/wrap (or otherwise adjust themselves to the available width)

**Additional context**
The problem occurs because in `_TreeViewItem.build` method it uses a `Row` widget to position the appropriate padding for the item + expand/collapse icon + item content, but without the item content being wrapped in an `Expanded` widget. It's not enough to have the item content itself be an `Expanded` widget because the item content is wrapped inside a `Padding` widget before being put into the row, so even if the item content is expanded to the right size (which can be shorter than what is needed by the text), the surrounding padding then causes it to overflow still. The solution is to have `_TreeViewItem.build` wrap the `Padding` widget that contains the `item.content` in an `Expanded` widget.

I'll submit a PR with the fix momentarily.

klondikedragon:
Fixes #269.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 Date format: DD/MM/YYYY
 
+## [3.x.x]
+
+- Fix overflow behavior for `TreeViewItem` ([#270](https://github.com/bdlukaa/fluent_ui/pull/270))
+
 ## [3.10.0] - Localization, Indicators, CommandBar and Flyouts - [02/04/2022]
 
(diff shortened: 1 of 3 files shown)


WinXaito (approved):
Not tested but the code seems ok.

Issue #205, opened 2022-02-25
Add PaneItemExpander to the NavigationPane


AKA-LinCoder:
hi ,when i look at Readme file in pub,i found that navigation pane can have tree style , but when i use it ,i can not find how to build it ,is it not supported yet? or i just not find the right way to achieve， hope i can get your answer

WinXaito:
I think you talk about that:

![sidebar_ms_expand](https://user-images.githubusercontent.com/8223773/155692628-8aef9335-850b-4776-885b-1750ea123380.gif)

That's currently not implemented (I took this capture from XAML Controls Gallery).

But I don't found that in Readme, maybe you confuse with TreeView ?

AKA-LinCoder:
> I think you talk about that:
> 
> ![sidebar_ms_expand](https://user-images.githubusercontent.com/8223773/155692628-8aef9335-850b-4776-885b-1750ea123380.gif)
> 
> That's currently not implemented (I took this capture from XAML Controls Gallery).
> 
> But I don't found that in Readme, maybe you confuse with TreeView ?

yes,this is what i want,thanks

AndreiLunga:
+1

omar0141:
When will it be in fluent ui

AKA-LinCoder:
nfg

PR #277, opened 2022-04-09
Improve NavigationView Top mode

bdlukaa:
#205 

- **BREAKING** Removed `ItemsExtension`, consequently removing the `getPaneItemsOffsets` method
- **BREAKING** `InheritedNavigationView.itemIndex` has been renamed to `InheritedNavigationView.currentItemIndex`
- Items in top mode now aren't rendered in a scrollable view. Instead, the `DynamicOverflow` widget, introduced on #232, is used to render the items

## Pre-launch Checklist


henry2man:
@bdlukaa A question regarding this component with ".top" setting. Is the position of the InfoBadge at the top right intentional? Could it be parameterized?

I am creating an email-like UI and the badges in that position are weird.

bdlukaa:
@henry2man I also think so. I believe some property like `InfoBadgePosition` can be added, in another PR after this one.

PR #325, opened 2022-05-11
Build on Flutter 3.0.0

h3x4d3c1m4l:
Fixes build on Flutter 3.0.0 including any new warnings. Basic functionality tested using my [mobile app](https://github.com/h3x4d3c1m4l/TPMS4All/tree/feature/first_startup_wizard).

I tried finding a way to have it compiling on both Flutter < 3.0.0 and Flutter 3.0.0 but it seems Dart does not support preprocessing like many C style languages... If anyone knows a way let me know and I'll happily adapt my PR.

Therefore this does break building on Flutter < 3.0.0.

## Pre-launch Checklist


bdlukaa:
If `?` is used where `!` previously were, there probably will be backwars compatibility

lib/src/controls/form/combo_box.dart
```diff
@@ -1072,16 +1072,16 @@ class _ComboboxState<T> extends State<Combobox<T>> with WidgetsBindingObserver {
       ),
     };
     focusNode!.addListener(_handleFocusChanged);
-    final FocusManager focusManager = WidgetsBinding.instance!.focusManager;
+    final FocusManager focusManager = WidgetsBinding.instance.focusManager;
```
h3x4d3c1m4l, line 1075: If I would replace the `!` by `?` then the type of the variable would be `FocusManager?` instead of `FocusManager`, right?. I don't think that's desirable as I would require lots of more changes...

bdlukaa: I believe it'd be something like

```suggestion
    // ignore: invalid_null_aware_operator
    final FocusManager focusManager = WidgetsBinding.instance?.focusManager!;
```

Altough, we make use of `TextSelectionControls`, in which its API has changed. Then, I believe it's not possible to keep the backwars compatibility :/

bdlukaa:
@h3x4d3c1m4l tests are failing :/

grahamsmith:
@bdlukaa - It might be wise given this was a major release to spin up a pre-release and new major version here to help with the compatibility? I saw similar approaches when moving to not null etc. 

Gives people chance to move over to Flutter 3.0 and then eventually it can be the main release etc.

Awesome work on this project. It looks and feels wonderful.

grahamsmith:
The failing tests look easier to fix - I could always PR into this PR 😂

bdlukaa (approved):

h3x4d3c1m4l:
I see you merged it already :) Nice!

Will open another one upgrade flutter_lints and to fix the 74 new warnings. Also I looked at the possibilities for backwards compatibility.

But unfortunately, as Dart does not have any C language style preprocessing and conditional imports do not seem to be suitable (they don't support Flutter version as condition), I can't find a way to have it backwards compatible.

bdlukaa:
@h3x4d3c1m4l don't worry about flutter_lints, I'm already on that

PR #369, opened 2022-05-23
NavigationPane positioning

Arenukvern:
## Pre-launch Checklist

With this PR I would like to add an option for NavigationPane right positioning.
Thank you for your time and review!

This doesn't include changes from https://github.com/bdlukaa/fluent_ui/pull/366#issue-1244324932

- [x] I have updated Example accordingly


bdlukaa:
I like the thing on the Settings screen :)

Altough, I don't like the result of it. Try the RTL mode (scroll to the bottom on settings), I believe that's how it should look like when the position is set. Instead, all the buttons aren't inverted as they should be. The animation is also off: if you check, the name (on the top left) is moved alonside the panel.

lib/src/controls/navigation/navigation_view/pane.dart
```diff
@@ -61,6 +61,11 @@ enum PaneDisplayMode {
   auto,
 }
 
+enum PanePosition {
+  left,
+  right,
+}
```
bdlukaa, line 67: This is nice, but doesn't work well in a directionality context. The available options should be `start` and `end`, so that it'd work well with rtl languages

Also, documentation for these is missing

lib/src/controls/navigation/navigation_view/pane.dart
```diff
@@ -100,6 +106,12 @@ class NavigationPane with Diagnosticable {
 
   final GlobalKey paneKey = GlobalKey();
 
+  /// The navigation pane position, that can be displayed on the
+  /// left side or right side.
+  ///
+  /// Works only with [PaneDisplayMode.open], [PaneDisplayMode.compact]
```
bdlukaa, line 112: I believe it should also work with minimal pane mode (maybe depends on #669)

bdlukaa (changes requested):

Arenukvern:
@bdlukaa Hello! Happy New Year) 
Sorry, I thought that this PR is closed as doesn't meet guidelines of MS Fluent UI:) and just wanted to update outdated code:).

The reason why I started making the right side position is because it is convenient mode (in my opinion) for Excel Addins - the web apps, that can be used inside excel - please see example in [this repository](https://github.com/xsoulspace/sheets_manager_excel_addin)

So if you think it is useful to include it to the your Fluent UI library please let me know, I will work to add RTL support and fixes for your review comments.

PR #465, opened 2022-08-13
Changed source code of second combobox

DobreRadu:


bdlukaa:
Why?

DobreRadu:
When you press source code to see the code,it's the same as the one above with colors green,it's not the one with cats.

PR #515, opened 2022-09-06
TreeView: Fix indentation, fix size of expand-icon's hitbox, add option to choose spacing of item content

robkri:
This PR contains some changes to TreeView that seemed usefull to me when I was using the widget.

- All items of the same depth level now have the same indentation. Before, only items with the same parent were aligned and there was an 8-pixel-offset between children of different parents when one of them contained expandable items and the other did not.
- The hitbox for the expand icon of each item now uses the item's full height and is three times wider than the actual icon. Before it was restricted to the icon's size (8px by 8px) which was not very comfortable to use. Now it corresponds to the implementation in the explorer of Windows 10/11. 
- Now `TreeView` offers the option to choose whether the items of a TreeView use more or less spacing. The examples shown in the Microsoft documentation use a wider spacing than the actual implementation uses for the explorer in Windows 10/11. Both options are available now, as shown in the attached image.
- The build method of TreeViewItem now contains some short comments to make it easier to find individual parts of each item like the selection checkbox or the expand icon.

## Pre-launch Checklist

- [ x] I have updated `CHANGELOG.md` with my changes  
- [x ] I have run "dart format ." on the project  
- [x ] I have added/updated relevant documentation
TreeView_spacing_examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 Date format: DD/MM/YYYY
 
+## [next]
+
+- `TreeView` updated: 
+  - All items of the same depth level now have the same indentation. Before, only items with the same parent were aligned.
+  - The hitbox for the expand icon of each item now uses the item's full height and is three times wider than the actual icon. This corresponds to the implementation in the explorer of Windows 10/11.
+  - You can now choose whether the items of a TreeView should use narrow or wide spacing. The examples shown in the Microsoft documentation use a wider spacing than the implementation used in the explorer of Windows 10/11.
+  - The build method of TreeViewItem now contains some short comments to make it easier to find individual parts of each item like the selection checkbox or the expand icon.
+
 ## [4.0.0-pre.4] - Almost there - [02/09/2022]
 
 - `DisableAcrylic` now fully disable transparency of its decendents `Acrylic`s ([#468](https://github.com/bdlukaa/fluent_ui/issues/468))
--- a/lib/src/controls/navigation/tree_view.dart
+++ b/lib/src/controls/navigation/tree_view.dart
@@ -2,14 +2,18 @@ import 'package:flutter/foundation.dart';
 
 import 'package:fluent_ui/fluent_ui.dart';
 
-const double _whiteSpace = 28.0;
+const double _whiteSpace = 8.0;
 
 /// Default loading indicator used by [TreeView]
-const Widget kTreeViewLoadingIndicator = SizedBox(
-  height: 12.0,
-  width: 12.0,
-  child: ProgressRing(
-    strokeWidth: 3.0,
+const Widget kTreeViewLoadingIndicator = Padding(
+  // Padding to make it the same width as the expand icon
+  padding: EdgeInsetsDirectional.only(start: 6.0, end: 6.0),
+  child: SizedBox(
+    height: 12.0,
+    width: 12.0,
+    child: ProgressRing(
+      strokeWidth: 3.0,
+    ),
   ),
 );
 
@@ -413,6 +417,7 @@ class TreeView extends StatefulWidget {
     this.itemExtent,
     this.addRepaintBoundaries = true,
     this.usePrototypeItem = false,
+    this.narrowSpacing = false,
   })  : assert(items.length > 0, 'There must be at least one item'),
         super(key: key);
 
@@ -478,8 +483,12 @@ class TreeView extends StatefulWidget {
   /// item to be the same size as the first item in the tree view.
   final bool usePrototypeItem;
 
+  /// Whether or not to have narrow spacing between the contents of each item.
+  /// The default value is `false`.
+  final bool narrowSpacing;
+
   @override
-  _TreeViewState createState() => _TreeViewState();
+  State<TreeView> createState() => _TreeViewState();
 
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@ Date format: DD/MM/YYYY
 
 ## [4.0.0-pre.4] - Almost there - [02/09/2022]
 
+- `TreeView` updateds:
```
bdlukaa, line 5: This should be moved out of `4.0.0-pre.4`, since it was already published

```suggestion
- `TreeView` updates: 
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -478,6 +483,10 @@ class TreeView extends StatefulWidget {
   /// item to be the same size as the first item in the tree view.
   final bool usePrototypeItem;
 
+  /// Whether or not to have narrow spacing between the contents of each item.
+  /// The default value is `false`.
+  final bool narrowSpacing;
+
   @override
   _TreeViewState createState() => _TreeViewState();
```
bdlukaa, line 491: Just for convenience

```suggestion
  State<TreeView> createState() => _TreeViewState();
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -727,50 +752,69 @@ class _TreeViewItem extends StatelessWidget {
...
                           ),
                         )
-                    else if (item._anyExpandableSiblings)
-                      // if some child items are expandable and others are not,
-                      // make sure that they line up vertically the same for the
-                      // same depth
+                    else
+                      // Add padding to make all items of the same depth level
+                      // have the same indentation, regardless whether or not
+                      // they are expandable.
```
bdlukaa, line 788: Can you explain this better? I don't get it

lib/src/controls/navigation/tree_view.dart
```diff
@@ -716,9 +738,12 @@ class _TreeViewItem extends StatelessWidget {
                 child: Row(
                   crossAxisAlignment: CrossAxisAlignment.center,
                   children: [
+                    // Checkbox for multi selection mode
                     if (selectionMode == TreeViewSelectionMode.multiple)
                       Padding(
-                        padding: const EdgeInsetsDirectional.only(end: 20.0),
+                        padding: EdgeInsets.only(
+                          right: narrowSpacing ? 0.0 : _whiteSpace,
+                        ),
```
bdlukaa, line 746: We must take directionality into consideration as well
```suggestion
                        padding: EdgeInsetsDirectional.only(
                          end: narrowSpacing ? 0.0 : _whiteSpace,
                        ),
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -2,14 +2,18 @@ import 'package:flutter/foundation.dart';
...
 
 /// Default loading indicator used by [TreeView]
-const Widget kTreeViewLoadingIndicator = SizedBox(
-  height: 12.0,
-  width: 12.0,
-  child: ProgressRing(
-    strokeWidth: 3.0,
+const Widget kTreeViewLoadingIndicator = Padding(
+  // Padding to make it the same width as the expand icon
+  padding: EdgeInsets.only(left: 6.0, right: 6.0),
```
bdlukaa, line 10:
```suggestion
  padding: EdgeInsetsDirectional.only(start: 6.0, end: 6.0),
```

lib/src/controls/navigation/tree_view.dart
```diff
@@ -688,12 +701,21 @@ class _TreeViewItem extends StatelessWidget {
...
+              // Indentation and selection indicator for single selection mode.
               Container(
                 height: selectionMode == TreeViewSelectionMode.multiple
                     ? 28.0
                     : 26.0,
                 padding: EdgeInsetsDirectional.only(
-                  start: 12.0 + item.depth * _whiteSpace,
+                  start: selectionMode == TreeViewSelectionMode.multiple
+                      ? !narrowSpacing
+                          ? item.depth * _whiteSpace * 3
```
bdlukaa, line 712: I believe we could add a top-level const for the narrow spacing factor. What do you think?

bdlukaa (changes requested):

lib/src/controls/navigation/tree_view.dart
robkri: I should have added a screenshot to illustrate:
TreeView_indentation
As you can see with the green line, all items with the same parent have the same indentation, no matter if they are expandable or not. But items of the same level, but with different parents, don't get the same indentation but show an offset (red line).

lib/src/controls/navigation/tree_view.dart
robkri: I'm not sure if I understand what you would like to do. 
```
selectionMode == TreeViewSelectionMode.multiple ?
   !narrowSpacing ? 
      item.depth * _whiteSpace * 3          // multi-select + wide spacing: add a padding of 24px per depth level
      :                                     // to have a little gap between the checkboxes
      item.depth * (_whiteSpace * 2 + 4)    // multi-select + narrow spacing: add a padding of 20px per depth level
   :                                        // to have the child's checkbox' start line up with the parent's checkbox' end
   !narrowSpacing ? 
      item.depth * _whiteSpace * 2          // single-select + wide spacing: add a padding of 16px per depth level
      :                                     // to have a 8px gap between the expand icons
      item.depth * _whiteSpace              // single-select + narrow spacing: add a  padding of just 8px per depth level
```

For which of these factors would you like to add a const?

lib/src/controls/navigation/tree_view.dart
bdlukaa: Got it!

bdlukaa (approved):

PR #521, opened 2022-09-08
Fix bug: The disabled TextBox blocks the focus traversal by pressing SHIFT+Tab keys

hnbsoft:
Fix bug: The disabled TextBox blocks the focus traversal by pressing SHIFT+Tab keys

* Bug reproduction steps:  
1. Start fluent_ui example in debug mode
2. Select TextBox example in the left sidebar.
3. Press Tab a few times to switch focus  
4. When the focus switches to the control after the "Disabled" TextBox
5. Then you press "SHIFT+TAB", the focus cannot be switched back to the control before the "Disabled" TextBox.

## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/form/text_box.dart
+++ b/lib/src/controls/form/text_box.dart
@@ -649,6 +649,7 @@ class _TextBoxState extends State<TextBox>
             : TextEditingValue(text: widget.initialValue!),
       );
     }
+    _effectiveFocusNode.canRequestFocus = enabled;
     _effectiveFocusNode.addListener(_handleFocusChanged);
   }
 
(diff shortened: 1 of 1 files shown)


lib/src/controls/form/text_box.dart
```diff
@@ -649,6 +649,7 @@ class _TextBoxState extends State<TextBox>
             : TextEditingValue(text: widget.initialValue!),
       );
     }
+    _effectiveFocusNode.canRequestFocus = enabled;
```
bdlukaa, line 652: this is unecessary since the same is implemented in `didChangeDependencies`

hnbsoft: Hello, I checked the flutter's TextField widget and it added this line, I'm not sure if this line is redundant

bdlukaa (approved):

PR #524, opened 2022-09-10
Expander header's height determination

sohaibbg:
The height of Expander's header is now determined based on the height of its children.

## Pre-launch Checklist


sohaibbg:
This was achieved by encasing the content widget in a Flexible widget so the Column can size the header as per its children's requirements.

lib/src/controls/surfaces/expander.dart
```diff
@@ -196,87 +190,87 @@ class ExpanderState extends State<Expander>
...
-              shape: widget.headerShape?.call(open) ??
-                  RoundedRectangleBorder(
-                    side: BorderSide(
-                      color: theme.resources.cardStrokeColorDefault,
-                    ),
-                    borderRadius: BorderRadius.vertical(
-                      top: const Radius.circular(4.0),
-                      bottom: Radius.circular(open ? 0.0 : 4.0),
-                    ),
+        builder: (context, states) => Container(
```
bdlukaa, line 193: use `return` instead of `=>`

lib/src/controls/surfaces/expander.dart
```diff
@@ -100,11 +99,6 @@ class Expander extends StatefulWidget {
   /// open and `false` when closed.
   final ValueChanged<bool>? onStateChanged;
 
-  /// The height of the header.
-  ///
-  /// Defaults to 48.0
-  final double headerHeight;
```
bdlukaa, line 106: Instead of removing this property, maybe we could change it to a `BoxConstraints`. What do you think?

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 Date format: DD/MM/YYYY
+## Expander height determination 10/09/2022
+- Determine the height of Expander's header from the height of its children.
+- Removed headerHeight parameter
+
```
bdlukaa, line 5:
```suggestion
## [next]
- **BREAKING** `Expander` header height is now dynamically calculated. Removed `Expander.headerHeight`

```

bdlukaa (changes requested):

lib/src/controls/surfaces/expander.dart
sohaibbg: sure, go ahead.

lib/src/controls/surfaces/expander.dart
sohaibbg: okay.

lib/src/controls/surfaces/expander.dart
sohaibbg: Actually I believe we should leave this to be determined by the passed header widget itself. If the user wants to pass advanced constraints, they can specify it in the header parameter.

bdlukaa:
Expander padding is off

example 12_09_2022 19_27_38

lib/src/controls/surfaces/expander.dart
sohaibbg: this has been implemented in the latest commit [54fde58]

lib/src/controls/surfaces/expander.dart
sohaibbg: The Container parent of header has been appended with a BoxConstraints property which is passed minHeight (default is 48.0 matching earlier headerHeight). minHeight is also now an optional parameter of Expander.

sohaibbg:
> Expander padding is off
> 
> example 12_09_2022 19_27_38

In latest commit, padding has been added to Container widget responsible for header.

lib/src/controls/surfaces/expander.dart
```diff
@@ -213,7 +213,11 @@ class ExpanderState extends State<Expander>
                     ),
                   ),
             ),
-            padding: const EdgeInsetsDirectional.only(start: 16.0),
+            padding: const EdgeInsetsDirectional.only(
+              start: 16.0,
+              top: 16.0,
+              bottom: 16.0,
```
bdlukaa, line 219: Since we now have `minHeight`, this turns out to be unecessary. It's also causing the expander to be extremely tall.

Fluent UI Showcase for Flutter e mais 1 página - Pessoal — Microsoft​ Edge 13_09_2022 19_12_07

Also, I believe it's causing `trailing` and `leading` to be small

lib/src/controls/surfaces/expander.dart
```diff
@@ -262,21 +266,25 @@ class ExpanderState extends State<Expander>
...
-      SizeTransition(
-        sizeFactor: _controller,
-        child: Container(
-          width: double.infinity,
-          padding: const EdgeInsets.all(16.0),
-          decoration: BoxDecoration(
-            border: Border.all(
-              color: theme.resources.cardStrokeColorDefault,
+      // CONTENT
+      Flexible(
```
bdlukaa, line 270: Why?

bdlukaa (changes requested):

lib/src/controls/surfaces/expander.dart
sohaibbg: The problem is in multi-line text. Perhaps top, bottom padding values of 8.0 would be more appropriate? Or I can remove them altogether as you say.

Frame 1

lib/src/controls/surfaces/expander.dart
sohaibbg: This enables us to put two expanded expanders in one constrained space, otherwise there is an issue. I'm trying to modify the code further to enable the same two expanders to then resize dynamically as they currently aren't doing so. I'll update here when I've worked that out. Thanks.

bdlukaa: Sure!

lib/src/controls/surfaces/expander.dart
bdlukaa: Well, to overcome this issue, I guess the padding must be applied to the header content itself. What do you think?

sohaibbg: That's actually a much better way to do things. I've Implemented this in the latest commit.

lib/src/controls/surfaces/expander.dart
sohaibbg: The work I was intending to do was too long and can be implemented in the future. As it stands, this change enables us to put ListViews directly in Expander contents. Its most useful where Expanders are meant to occupy the full screen space. I would say this commit is final then and you can see if it meets the requirements to be merged. Thanks!

bdlukaa:
The trailing is being constrained down, for some reason

image

PR #531, opened 2022-09-14
Revert attribute content and add assert with pane attribute

WinXaito:
## Pre-launch Checklist

I have some case where I use the navigation view like this:

```dart
return NavigationView(
  content: ScaffoldPage(
    header: PageHeader(
      title: titleRow,
    ),
    content: Padding(
      padding: const EdgeInsets.symmetric(horizontal: 25),
      child: child,
    ),
  ),
);

And after the PR #510 this is not possible, we need to do something like that:

```dart
return NavigationView(
  pane: NavigationPane(
      displayMode: PaneDisplayMode.minimal,
      items: [
        PaneItem(
          icon: const SizedBox.shrink(),
          body: ScaffoldPage(
            header: PageHeader(
              title: titleRow,
            ),
            content: Padding(
              padding: const EdgeInsets.symmetric(horizontal: 25),
              child: child,
            ),
          ),
        )
      ]
  ),
);
```

With this fix, we can use one solution with `pane` or `content` (but one of these must be null !).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@ Date format: DD/MM/YYYY
 
 ## [next]
 
-- **BREAKING** Removed `NavigationBody`. Use `PaneItem.body` instead ([#510](https://github.com/bdlukaa/fluent_ui/pull/510)):
+- **BREAKING** Removed `NavigationBody`. Use `PaneItem.body` instead ([#510](https://github.com/bdlukaa/fluent_ui/pull/510)/[#531](https://github.com/bdlukaa/fluent_ui/pull/531)):  
   Before:
   ```dart
-  NavigationBody(
+  NavigationView(
     pane: NavigationPane(
       items: [
         PaneItem(icon: Icon(FluentIcons.add)),
@@ -23,9 +23,9 @@ Date format: DD/MM/YYYY
   ),
   ```
   
-  Now:
+  Now:  
   ```dart
-  NavigationBody(
+  NavigationView(
     ...
     pane: NavigationPane(
       items: [
@@ -45,6 +45,19 @@ Date format: DD/MM/YYYY
     ),
   ),
   ```
+  
+  Or if you didn't have panes, you can use the content like that:  
+  ```dart
+  NavigationView(
+    content: ScaffoldPage(
+      header: PageHeader(
+        title: titleRow,
+      ),
+      content: child,
+    ),
+  ),
+  ```
+  *Note: one attribute of pane or content attribute must be null*
 
   You can use `NavigationView.transitionsBuilder`
 - Fixes memory leaks on `NavigationView`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ Date format: DD/MM/YYYY
     ),
   ),
   ```
-  *Note: one attribute of pane or content attribute must be null*
+  *Note: one attribute of pane or content must be null*
 
   You can use `NavigationView.transitionsBuilder`
 - Fixes memory leaks on `NavigationView`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ Date format: DD/MM/YYYY
   ),
   ```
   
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -45,6 +45,19 @@ Date format: DD/MM/YYYY
     ),
   ),
   ```
+  
+  Or if you didn't have panes, you can use the content like that:
```
bdlukaa, line 49:
```suggestion
  Or if you don't have panes, you can use the content like the following:  
```

lib/src/controls/navigation/navigation_view/view.dart
```diff
@@ -34,11 +38,14 @@ class NavigationView extends StatefulWidget {
...
     this.appBar,
     this.pane,
+    this.content,
     this.clipBehavior = Clip.antiAlias,
     this.contentShape,
     this.onOpenSearch,
     this.transitionBuilder,
-  }) : super(key: key);
+  })  : assert((pane != null && content == null) ||
+            (pane == null && content != null)),
```
bdlukaa, line 47: Add an error message:

```suggestion
  })  : assert(
    (pane != null && content == null) || (pane == null && content != null),
    'Either pane or content must be provided',
    ),
```

lib/src/controls/navigation/navigation_view/view.dart
```diff
@@ -47,6 +54,11 @@ class NavigationView extends StatefulWidget {
   /// left, on the top, or above the body.
   final NavigationPane? pane;
 
+  /// The content of the pane.
+  ///
+  /// Usually an [NavigationBody].
```
bdlukaa, line 59:
```suggestion
  /// The content of the pane.
  ///
  /// If [pane] is provided, this is ignored
  ///
  /// Usually a [ScaffoldPage]
```

lib/src/controls/navigation/navigation_view/view.dart
```diff
@@ -591,10 +603,10 @@ class NavigationViewState extends State<NavigationView> {
...
           }
         }
-      } else {
+      } else if (widget.content != null) {
         paneResult = Column(children: [
           appBar,
-          // Expanded(child: widget.content),
+          Expanded(child: widget.content!),
         ]);
       }
```
bdlukaa, line 611: We would never reach this, since either pane or content must be provided, but it's good to add some safety.

```suggestion
      } else {
        throw 'Either pane or content must be provided';
      }
```

bdlukaa (changes requested):

henry2man:
Prior to `4.0.0 stable` I had both `PaneItems` , a `NavigationPane.onChanged` simple callback and a `LazyIndexedStack` in content attribute: 

```dart
var index = 0;
...
NavigationView ( 
 content: LazyIndexedStack(
    index: index,
    children: [
                  (context) => const ScreenOne(),
                  (context) => const ScreenTwo(),
                  (context) => const ScreenThree(),
    ]),
pane: NavigationPane(
                    selected: index,
                    onChanged: (i) {
                      setState(() => index = i);
                    },
                    items: [
                      PaneItem(
                        icon: Icon(
                          Icons.home,
                        ),
                        title: Text("One"),
                      ),
                      PaneItem(
                        icon: Icon(
                          Icons.home,
                        ),
                        title: Text("Two"),
                      ),
                      PaneItem(
                        icon: Icon(
                          Icons.home,
                        ),
                        title: Text("Three"),
                      ),
```

With this setup I was able to maintain internal state for every page in the `LazyIndexedStack`. But since I've updated to `+4.0.0 stable` I had to decide between `PaneItem + body` or not having `PaneItems` at all. I could go for `PaneItem + body` if there were a mechanism  to preserve state, but until now `LazyIndexedStack` was the only option for me.

bdlukaa:
you can use a `NavigationView.paneBodyBuilder`

henry2man:
Sadly, `NavigationView.paneBodyBuilder` doesn't works (even with keys) because on every PaneItem page switch the state is rebuilt.

I've opened https://github.com/bdlukaa/fluent_ui/pull/595 in order to allow current behavior and pre `4.0.0` stable behavior, too

PR #551, opened 2022-09-30
Fix unnecessary NavigationAppBar leading icon when no NavigationPane was provided

DMouayad:
## Description
After updating from version 3.11 to 4.0.1 I encountered a problem with `NavigationAppBar`.
For screen width <= 640 it was showing an icon for openin'/closing the pane which is the expected behavior, but the issue is I did not provide a `NavigationPane` for the `NavigationView` so the icon is unnecessary.

I found in the source code that an additional leading icon is added to the app bar but without checking whether there's a pane or not.  
What I did basically is adding an If statement before adding the additional leading.

## Pre-launch Checklist

- [✔] I have updated `CHANGELOG.md` with my changes  
- [✔] I have run "dart format ." on the project


bdlukaa (approved):

bdlukaa (approved):
LGTM.

_I just updated the changelog, hope you don't mind_

Issue #548, opened 2022-09-28
make PaneItem.body is not required and make the pane and the content can be exist on one time


frg2089:
Hi.

I have a code like this
```dart
NavigationView(
  // ...
  pane: NavigationPane(
    selected: index,
    onChanged: _changedPage,
    items: [
      PaneItem(icon, title),
      PaneItem(icon, title),
      PaneItem(icon, title),
      if (_special) ...[
        PaneItemSeparator(),
        PaneItem(icon, title)
      ],
    ],
  ),
  content: Navigator(),
);
```

I can used `onChanged` event to push a page into `Navigator` at old version. (4.0.0-pre.4)

However, I cannot use it on latest version. (4.0.0 or higher)
Because at the latest version, the `pane` and the `content` cannot be exist on one time.

could you can make `PaneItem.body` is not required and make the `pane` and the `content` can be exist on one time?

Thanks.

klondikedragon:
The new behavior that makes `PaneItem.body` required also makes the way I'm using `NavigationPane` more difficult as well, where the child that is passed in comes from the parent width (vRouter with TabbedNester routes): https://github.com/bdlukaa/fluent_ui/issues/226#issuecomment-1070267070

In this router approach, the child widget for each page are specified alongside the routes, and when there are nested tabbed routes, the child for a given pane item can be dynamic based on a specific tabbed sub-route, so it's not obvious how to convert to the new style.

For a simple application, `PaneItem.body` is certainly more convenient, so it would be really helpful to support both modes. Perhaps `PaneItem.body` should still be required so that users of the simple case remember to do the right thing (easy enough to pass a `const SizedBox.shrink()` that would get ignored), but that if `content` is specified it will override any content that would have been built from the pane item body. I'll submit a PR shortly along this direction.

klondikedragon:
It looks like a cleaner approach will be an optional builder that can customize the NavigationBody built from the currently selected pane item, which could choose to just ignore the pane item's body and return something else instead.

Issue #226, opened 2022-03-14
🐛 Combining FluentApp.router() and NavigationPane doesn't animate properly


anaisbetts:
**Describe the bug**

Trying to use NavigationPane along with routing (e.g. to make each PaneItem a route) doesn't work for a number of reasons:

1. Default transitions make for a very weird, non-Windows experience. This can be fixed, but...
2. Navigating to a new page results in the NavigationPane playing its intro transition which looks Odd

**To Reproduce**
https://github.com/arquivolta/desktop/tree/navigation-bug demonstrates this issue, the Web driver works without any OS-specific dependencies

**Expected behavior**
My initial goal was to expose NavigationPane pages as routes and have clicking on the panes navigate, but it appears that this isn't really possible, and I'll only be able to use Routing for e.g. dialogs or intro pages. 

I may be Doing It Wrong, it would be cool to see an example of combining Fluent UI and routing in the way that Fluent UI expects

klondikedragon:
Not sure how to do it with the Beamer router, but I currently have navigation working well (at least on web, haven't tested on other platforms) in conjunction with [vRouter](https://github.com/lulupointu/vrouter). I don't have any animation when it switches between panes (whether the switch was caused by a route change or the user clicking on a navbar pane item), which feels more "Windows-like" to me.

High level strategy:
* Using TabbedNester code from https://github.com/lulupointu/vrouter/pull/176 to more easily create nested routes (with the nested routes being one route per pane)
* It uses an IndexedStack so that the state of widgets on non-visible panes is preserved
* It uses some state to remember the last nested sub-route on a given pane

Top-level app is building something like this:
<details>
<summary>Show code</summary>

```dart
              VRouterScope(
                child: FluentApp.router(
                  ...
                  routeInformationParser: VRouteInformationParser(),
                  routerDelegate: getRouterDelegate(ref),
                  backButtonDispatcher: VBackButtonDispatcher(),
                ),
              )
```
</details>

And then the `getRouterDelegate` function has a structure like...
<details>
<summary>Show code</summary>

```dart
VRouterDelegate getRouterDelegate(WidgetRef ref) {
    // Make sure that if we have no current workspace that we
    // enforce the routing rules.
    final _ = ref.watch(guardWorkspacesNavProvider);
    return VRouterDelegate(
      navigatorKey: locator<NavigatorService>().navigatorKey,
      routes: [
                    TabbedNester(
                      path: "/",
[...]

PR #553, opened 2022-10-01
Add NavigationView.paneBodyBuilder

klondikedragon:
Useful to override the widget used as the body of the navigation view in conjunction with pane items. This can be useful if you are using router-based navigation, and the body of the navigation pane should be determined by the current route rather than just by the currently selected pane.

This resolves https://github.com/bdlukaa/fluent_ui/issues/548.

It allows 4.0.x to be compatible with the approach described in https://github.com/bdlukaa/fluent_ui/issues/226#issuecomment-1070267070 with the following modification to `MyAppNavBar` widget:

<details>
<summary>Example use of paneBodyBuilder</summary>

```
...
class _MyAppNavBarState extends ConsumerState<MyAppNavBar> {
  // Will be populated on demand as each of the navbar panes is visited (if ever).
  List<Widget> panes = List.filled(topNavPages.length, Container());
  // Track the last specific route we were using within a given nav pane, so that
  // we can restore the full sub-route when we go back to that particular nav pane.
  List<String> paneLastVisitedUrls = List.from(topNavPages);

  @override
    // Populate current widget and route for the currently selected pane.
    panes[widget.currentTabIndex] = widget.child;
    paneLastVisitedUrls[widget.currentTabIndex] = context.vRouter.url;

    return material.Scaffold(
        body: NavigationView(
      pane: NavigationPane(
        selected: (widget.currentTabIndex >= 0 &&
                widget.currentTabIndex < numPaneItems)
            ? widget.currentTabIndex
            : null,
        onChanged: (i) => context.vRouter.to(paneLastVisitedUrls[i]),
        // TODO: Watch this until https://github.com/bdlukaa/fluent_ui/issues/173 is fixed
        indicatorBuilder: NavigationIndicator.sticky,
        //size: const NavigationPaneSize(),
        header: ...,
        displayMode: PaneDisplayMode.compact,
        items: [
            PaneItem(icon: ..., Text('Home'), body: const SizedBox.shrink()),
            PaneItem(icon: ..., Text('Pane 2'), body: const SizedBox.shrink()),
            PaneItem(icon: ..., Text('Pane 3'), body: const SizedBox.shrink()),
            PaneItem(icon: ..., Text('Pane 4'), body: const SizedBox.shrink()),
        ],
        footerItems: [
          PaneItemSeparator(),
          PaneItem(icon: ..., Text('Pane 5'), const SizedBox.shrink()),
        ],
      ),
      // Use IndexedStack instead of NavigationBody so that the state
      // of the children pages are preserved between navigations
      paneBodyBuilder: (viewWidget, selectedIndex, selectedPaneItemBody) =>
          NavigationBody(
        itemKey: ValueKey(selectedIndex ?? -1),
        transitionBuilder: viewWidget.transitionBuilder,
        child: IndexedStack(
          index: widget.currentTabIndex,
          children: panes,
        ),
      ),
    ));
  }
}
```
</details>

## Pre-launch Checklist


bdlukaa:
I don't like the design of this implementation. `NavigationBody` was made private to make the animation consistent with other properties/features of `NavigationView`. If we'd have dynamic content, I'd make a `Widget content` property, which would show the current body.

klondikedragon:
@bdlukaa - I've refactored, see if you like this second approach better. It makes `_NavigationBody` private again and doesn't require the builder callback to know anything about its internals, but it still allows the dynamic pane content to benefit from the animation provided by it.

Another option would be to just have it pass through and use the content widget if it's non-null in the place of pane.selectedItem.body. That would also work for me, but perhaps be a little less flexible for other use cases that might want to dynamically adapt the contents of the current pane item body. I'm happy to change the PR to this third approach if you prefer it.

bdlukaa (approved):

PR #556, opened 2022-10-02
TreeView selection state fixes

### LINKED ISSUE #555: 🐛 TreeView selection state fixes/enhancements
All the recent enhancements to `TreeView` are very much appreciated, especially the `narrowSpacing` behavior and the keyboard navigation!

My application is using both single-selection and multi-selection `TreeView` widgets, and I've noticed some bugs and useful enhancements:
* 🐛 A single-selection `TreeView` that is given an initial list of items that has a single item selected may improperly clear out the selection state of this item if the single selected item has any children (`initState` calls `buildItems` and that improperly calls `updateSelected`).
* 🐛 In single-selection mode, if a child item is selected, and then the parent is collapsed so that the child is hidden, and then the parent item is selected, the child item's selected state is not properly set to false. If you expand the parent, then the child item will still be selected and it will have multiple items selected even though it's a single-selection tree view.
* `onItemInvoked` is called for 3 different reasons: when the item's selection state is toggled, when the item is expanded or collapsed, or when the item is otherwise pressed (or space/enter is pressed). Without knowing the item invoke reason, it's not easy to implement logic that toggles the selection state of an item only when it's pressed (and not for the other two reasons).
* Along the same lines, an application may want to toggle the selection state of a given item in a multi-selection mode tree, and some helper methods should be exposed so that the same logic can be used to set the new selection state of an item and all of its children, as well as to appropriately update the selection state of the parent(s) of an item.
* You may want to allow a parent item to remain selected even if the user deselects all of the children (it would show the "indeterminate" checkbox state in this case). This can be useful if the parent
* For this reason, you may also want to have included in the list of `selectedItems` passed to the `onSelectionChanged` callback any parent items that are partially selected.

I'll submit a PR that fixes these bugs and implements these enhancements. With these enhancements, my application can now easily programmatically change the selection state when an item is clicked/pressed (storing selection state in riverpod and propagating changed item list back to treeview widget through provider rebuild), and also parent items can remain selected even if children are all deselected. Note that it still defaults to the old behavior.

klondikedragon:
This fixes/implements enhancements in #555:
  - **BREAKING** Added `TreeViewItemInvokeReason` parameter to `TreeView.onItemInvoked` and `TreeViewItem.onInvoked`.
  - Fix clearing out selection state on initial state build in certain cases for a single selection mode tree view.
  - Fix single selection mode to properly deselect hidden child items when selecting a collapsed parent item.
  - Add `TreeView.includePartiallySelectedItems` so that items who have children with a mixed selection state will be included in the `onSelectionChanged` callback.
  - Add `TreeView.deselectParentWhenChildrenDeselected` optional behavior so that parent items can remain selected when all of their children are deselected.
  - Add `TreeViewItem.setSelectionStateForMultiSelectionMode` helper method and `[TreeViewItem].selectedItems` extension method, to make it easier for application code to programmatically change selection state of items in a multi-selection mode tree view.

Resolves #555.

## Pre-launch Checklist


bdlukaa (approved):
LGTM!

_I just updated the changelog, hope you don't mind_

PR #597, opened 2022-11-05
Fix TreeView selection state for unexpanded items

### LINKED ISSUE #578: TreeView initial value on selected.
Is there a way that the parent will has an initial value also that depends on the children?

Screen Shot 2022-10-20 at 7 52 58 PM

klondikedragon:
Resolves #578

## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/navigation/tree_view.dart
+++ b/lib/src/controls/navigation/tree_view.dart
@@ -276,7 +276,7 @@ class TreeViewItem with Diagnosticable {
     bool hasFalse = false;
     bool hasTrue = false;
 
-    for (final child in children.build(assignInternalProperties: false)) {
+    for (final child in children) {
       if (child.selected == null) {
         hasNull = true;
       } else if (child.selected == false) {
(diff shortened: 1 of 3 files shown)


lib/src/controls/navigation/tree_view.dart
```diff
@@ -276,7 +276,7 @@ class TreeViewItem with Diagnosticable {
     bool hasFalse = false;
     bool hasTrue = false;
 
-    for (final child in children.build(assignInternalProperties: false)) {
```
bdlukaa, line 279: Why don't we build the children here?

klondikedragon: The `build` method only returns visible children, and the selection state also needs to consider children that are not visible because they are underneath a non-expanded item. Also, updateSelected is being called on all parent nodes for any `TreeViewItem` whose selection state has changed in a depth-first fashion (or in the case of the initial load, for all items). If `updateSelected` recursively builds all children again, it will make the traversal cost of updating selection state O(n^2) in the size of the tree rather than O(n). This could be a big difference for really large trees.

I've added some more documentation to `updateSelected` to indicate the proper way to call it, for future references. All current callers already do follow this pattern.
```dart
item.executeForAllParents((parent) => parent
  ?.updateSelected(widget.deselectParentWhenChildrenDeselected))
```

bdlukaa (approved):

Issue #678, opened 2023-01-14
🐛 Assertion triggered if nav pane selection changes and there is no selected item


klondikedragon:
**Describe the bug**
In some of my pane items for `NavigationPane`, I'm using a `PaneItemAction` that will show a popup menu when pressed, and some of the action items on that menu will navigate to a page (using a router) that results in no top-level navbar item being selected. i.e. they're navigating to some area of the app that has no corresponding top-level navbar item.

There was a [recent change](https://github.com/bdlukaa/fluent_ui/commit/520527d634128136f92daa371ee211c2b95a82da#diff-0713dcda8014602ee23c537675802acf6a0611aee19fbb69610a3f106fb76f6aR218) where that doesn't properly consider that the selected item may be null. In the statement:
```dart
final item = widget.pane?.selectedItem.itemKey.currentContext;
```

the call to the `selectedItem` getter will assert if there is no selected item. (The code after that line does consider that `item` itself may be null.) So this should be changed to something like:

```dart
final item = widget.pane?.selected == null ? null : widget.pane?.selectedItem.itemKey.currentContext;
```

PR #680, opened 2023-01-14
Allow paneBodyBuilder to use an IndexedStack without interference

### LINKED ISSUE #679: 🐛 interference in _NavigationBodyState between PageView and paneBodyBuilder/IndexedStack
**Describe the bug**
I've recently upgraded fluent_ui from 4.0.1 to 4.1.4, and started to receive all kinds of odd/random issues. e.g., duplicate GlobalKeys in widgets related to a nested vrouter, stateful widgets being disposed almost immediately after creation (causing futures executed in a "zero" delay created in an initState that reference the build context to get an exception that the widget was already detached), widgets not having a sized viewport during construction, etc.

It appears this is related to the recent enhancements (#607, #645) surrounding preserving state of navigation pane items through the ability to specify a key for each pane item. This is a great enhancement and good when the use of navigation pane is simple (common use case for most apps probably). In my case, I need to manage the keys/state of the widgets presented within the body of the navigation pane explicitly (and underneath an `IndexedStack`) because the content is not determined just by the currently selected pane, but also by the current route, which could include sub-routes that reference tabs with different content. (A sketch of what is being done is in https://github.com/bdlukaa/fluent_ui/issues/226#issuecomment-1272350076 ).

It appears that using an IndexedStack in combination with `paneBodyBuilder` is contemplated (somewhat discussed in https://github.com/bdlukaa/fluent_ui/issues/649#issuecomment-1372634025).

I'm not exactly sure why, but the new `PageView.builder` (with an automatic GlobalKey passed to it) is really interfering strongly with the `IndexedStack` returned by my `paneBodyBuilder`.
 
I've tested a solution that seems to work nicely, where if the `paneBodyBuilder` is not null then it just doesn't wrap the widget built by the `paneBodyBuilder` in the `PageView.builder` (nor does it use the `_pageKey` in `_NavigationBodyState`) and that removes all of the problems.

I've created a PR for consideration to fix this issue and #678.

klondikedragon:
Fixes #679, as well as the minor assertion in #678.

### DIFF
--- a/lib/src/controls/navigation/navigation_view/body.dart
+++ b/lib/src/controls/navigation/navigation_view/body.dart
@@ -140,27 +140,39 @@ class _NavigationBodyState extends State<_NavigationBody> {
             child: child,
           );
         },
-        child: KeyedSubtree(
-          key: widget.itemKey,
-          child: PageView.builder(
-            key: _pageKey,
-            physics: const NeverScrollableScrollPhysics(),
-            controller: pageController,
-            itemCount: view.pane!.effectiveItems.length,
-            itemBuilder: (context, index) {
-              final isSelected = view.pane!.selected == index;
-              final item = view.pane!.effectiveItems[index];
+        child: () {
+          final paneBodyBuilder = widget.paneBodyBuilder;
+          if (paneBodyBuilder != null) {
+            return FocusTraversalGroup(
+              key: widget.itemKey,
+              child: paneBodyBuilder.call(view.pane?.selected != null
+                  ? view.pane?.selectedItem.body
+                  : null),
+            );
+          } else {
+            return KeyedSubtree(
+              key: widget.itemKey,
+              child: PageView.builder(
+                key: _pageKey,
+                physics: const NeverScrollableScrollPhysics(),
+                controller: pageController,
+                itemCount: view.pane!.effectiveItems.length,
+                itemBuilder: (context, index) {
+                  final isSelected = view.pane!.selected == index;
+                  final item = view.pane!.effectiveItems[index];
 
-              return ExcludeFocus(
-                key: item.bodyKey,
-                excluding: !isSelected,
-                child: FocusTraversalGroup(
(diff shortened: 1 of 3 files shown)


klondikedragon:
Note that this has been tested not to change the previous behavior if `paneBodyBuilder` is null, and also to fully fix all of the compatibility issues if `paneBodyBuilder` is specified and returns an `IndexedStack` widget (as for example used in https://github.com/bdlukaa/fluent_ui/issues/226#issuecomment-1272350076).

lib/src/controls/navigation/navigation_view/body.dart
```diff
@@ -140,27 +140,39 @@ class _NavigationBodyState extends State<_NavigationBody> {
...
-        ),
+        child: () {
+          final paneBodyBuilder = widget.paneBodyBuilder;
+          if (paneBodyBuilder != null) {
+            return FocusTraversalGroup(
+              key: widget.itemKey,
+              child: paneBodyBuilder.call(view.pane?.selected != null
+                  ? view.pane?.selectedItem.body
+                  : null),
+            );
```
bdlukaa, line 151: `itemKey` should be for every item. What do you think?

klondikedragon: The code path where `paneBodyBuilder` is non-null essentially reverts to the [old behavior](https://github.com/bdlukaa/fluent_ui/pull/607/files#diff-b33184314dc53c5740bf048c731cfa0f34c0d6e05b02311863562e81ef7d5387R137). It looks like the itemKey being passed in to `_NavigatorBody` has the value of ```itemKey: ValueKey(pane.selected ?? -1),```, so this would be having the `FocusTraversalGroup` be differentiated for each group of controls separately for each pane item (what we want)?

I tried removing the `key` from the `FocusTraversalGroup` entirely and it didn't seem to break anything, so I'm not sure why it's even needed at all here.

Issue #622, opened 2022-11-24
🐛 FlyoutPlacement.full doesn't stretch the Flyout to fill the whole window


robkri:
**Describe the bug**
When using `FlyoutPlacement.full` on a `Flyout`-widget, the documentation claims that "The flyout will be streched and positioned on the whole app window." But apparently the Flyout (blue) only stretches to fill the app window _above_ the child and there seems to be no way to use `FlyoutPlacement.full` and position the Flyout below the child.

![Flyout_FlyoutPlacement_full](https://user-images.githubusercontent.com/103884835/203757166-b7c9a249-dfea-4a30-aa24-ad4042896a82.jpg)

**To Reproduce**
```
Center(
  child: Flyout(
    openMode: FlyoutOpenMode.press,
    placement: FlyoutPlacement.full,
    position: FlyoutPosition.below,      // This seems to have no effect as soon as FlyoutPlacement.full is set.
    content: (context) => Container(
      color: Colors.blue,
      width: double.infinity,
      height: double.infinity,
      child: const Align(
        alignment: AlignmentDirectional.bottomEnd,
        child: FlyoutContent(
          child: SizedBox(child: Text("FlyoutContent")),
        ),
      ),
    ),
    child: Container(
      color: Colors.orange,
      child: const Padding(
        padding: EdgeInsets.all(8.0),
        child: Text("Click to open Flyout"),
      ),
    ),
  ),
)
```

**Expected behavior**
The `Flyout` stretches to fill the entire app window.

PR #690, opened 2023-01-21
Flyout rework

### LINKED ISSUE #27: Implement menus and context menus
Menus and context menus display a list of commands or options when the user requests them. Use a menu flyout to show a single, inline menu. Use a menu bar to show a set of menus in a horizontal row, typically at the top of an app window. Each menu can have menu items and sub-menus.

Anatomy:
- MenuBar
       - MenuBarItem (Essentially a button)
              - MenuFlyoutItem (Essentially a tappable list tile)

The animation is a `SizeTransition` combined with ease curve and faster duration.

bdlukaa:
Fixes #27

--- 
  Flyouts were reworked to match the design and behavior of native WinUI 3

  **BREAKING** Removed `Flyout` widget.
  To replace it, `FlyoutTarget` and `FlyoutController` were created. `FlyoutTarget` works like a target, which the given `controller` will use to display the flyout

  ---
  Migration guide:

  Before:
  ```dart
  final controller = FlyoutController();

  Flyout(
    controller: controller,
    placement: ...,
    position: ...,
    verticalOffset: ...,
    onOpen: ...,
    onClose: ...,
    child: Button(
      onPressed: controller,
      child: Text('Tap me'),
    ),
  ),
  ```

  Now:
  ```dart
  final controller = FlyoutController();

  FlyoutTarget(
    controller: controller,
    child: Button(
      onPressed: _showFlyout,
      child: Text('Tap me'),
    ),
  ),

  void _showFlyout() async {
    await controller.showFlyout(
      barrierDismissible: ...,
      dismissWithEsc: ..., // NEW
      dismissOnPointerMoveAway: ..., // NEW
      placementMode: ...,
      autoModeConfiguration: ..., // NEW
      forceAvailableSpace: ..., // NEW
      shouldConstrainToRootBounds: ..., // NEW
      additionalOffset: ...,
      margin: ..., // NEW
      barrierColor: ...,
      navigatorKey: ..., // NEW
      transitionBuilder: ..., // NEW
      transitionDuration: ..., // NEW
      builder: (context) => FlyoutContent(...),
    ); 
  }
  ```
  ---

  Now, it's possible to dismiss the flyout by tapping the barrier (`barrierDismissible`), pressing the `ESC` keyboard key (`dismissWithEsc`) and by moving the cursor (pointer) away from the flyout (`dismissOnPointerMoveAway` - defaults to false);

  Automatic mode is finally implemented, and it's the default mode. By setting `autoModeConfiguration`, it's possible to customize the preferred flyout placement. If flyout doesn't meet the placement conditions, it decides where it fits the best.

  `forceAvailableSpace` determines whether the flyout size should be forced the available space according to the attached target. It's useful when the flyout is large but can not be on top of the target. `NavigationView`'s top navigation mode now uses it by default on pane items overflow.

  `shouldConstrainToRootBounds` determines whether the flyout should fit the root bounds - usually the window bounds. If false, the flyout will be able to overflow the screen on all sides. Defaults to `true`

  `margin` determines the margin of the flyout to the root. `additionalOffset` determines the margin of the flyout to the target.

  It's now possible to assign a custom transition to the flyout by providing `transitionBuilder` and `transitionDuration`. By default, a light slide-fade transition is used, but it can be highly customizable to fit your needs. It provides the current placement mode - since automatic mode may change it at layout time. `DropdownButton` uses it to create its slidethrough transition.

  `position` and `placement` were replaced by `placementMode`, which gives horizontal and vertical options of placement, at all screen alignments. It's also possible to use it in a right-to-left context by using `placementMode.resolve(Directionality.of(context))`

## Pre-launch Checklist


bdlukaa:
Some updates to `FlyoutContent` and `FlyoutListTile` are still necessary until this becomes ready. Some tests would be nice as well

bdlukaa:
There should be an option to position the flyout anywhere in the screen (https://github.com/bdlukaa/fluent_ui/issues/622)

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+
+  final VoidCallback onDismiss;
+
+  @override
+  void invoke(covariant DismissIntent intent) {
+    onDismiss();
+  }
+}
+
+class FlyoutAttach extends StatefulWidget {
```
HrX03, line 739: `FlyoutTarget` instead maybe? Sounds better, FlyoutAttach doesn't really tell you clearly what it's meant for

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+    Key? key,
+    required this.controller,
+    required this.child,
+  }) : super(key: key);
+
+  @override
+  State<FlyoutAttach> createState() => FlyoutAttachState();
+}
+
+class FlyoutAttachState extends State<FlyoutAttach> {
```
HrX03, line 753: Is there any reason for this to be public? The controller seems to live on the same file so

lib/src/controls/flyouts/content.dart
```diff
@@ -18,54 +17,76 @@ class FlyoutContent extends StatelessWidget {
...
+  ///
+  /// If null, [ThemeData.menuColor] is used by default
   final Color? color;
 
   /// The shape to fill the [color] of the box.
   final ShapeBorder? shape;
 
   /// Empty space to inscribe around the [child]
+  ///
+  /// Defaults to 8.0 in all sides
```
HrX03, line 36: nit: on each side instead of in all sides?

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
+import 'package:fluent_ui/fluent_ui.dart';
+
+class ContentManager extends StatefulWidget {
```
HrX03, line 3: Is this documented anywhere?

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+  State<ContentManager> createState() => _ContentManagerState();
+}
+
+class _ContentManagerState extends State<ContentManager> {
+  final GlobalKey key = GlobalKey();
+  Size size = Size.zero;
+
+  @override
+  void initState() {
+    WidgetsBinding.instance.addPostFrameCallback((timeStamp) {
```
HrX03, line 21: note: addPostFrameCallback makes the framework "lose" a frame as it needs to rebuild two times and as such it's not recommended. Is there really no other way to get the size from context without this? Maybe from within a callback or such

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return ContentSizeInfo(
+      size: size,
+      child: Builder(builder: (context) {
+        return KeyedSubtree(
+          key: key,
+          child: widget.content(context),
+        );
+      }),
```
HrX03, line 39: nit: trailing comma

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+    );
+  }
+}
+
+class ContentSizeInfo extends InheritedWidget {
+  const ContentSizeInfo({
```
HrX03, line 45: Consider making the constructor private if you're not planning to expose this outside of this file

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+}
+
+typedef _MenuBuilder = Widget Function(
+  BuildContext context,
+  BoxConstraints rootSize,
+  Iterable<Widget> menus,
+  Iterable<GlobalKey> keys,
+);
+
+@protected
```
HrX03, line 72: This annotation here makes no sense, works only on instance members, not on classes themselves

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+    Key? key,
+    required this.builder,
+    required this.flyoutKey,
+    required this.additionalOffset,
+    required this.margin,
+    required this.transitionDuration,
+  }) : super(key: key);
+
+  final GlobalKey flyoutKey;
+  final _MenuBuilder builder;
```
HrX03, line 84: Exposing a private typedef by using a public class is rather weird as you can't directly reference said typedef but must copy the entire signature from the source file, consider making it public

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  static MenuInfoProviderState of(BuildContext context) {
+    return context.findAncestorStateOfType<MenuInfoProviderState>()!;
+  }
+
+  @override
+  State<MenuInfoProvider> createState() => MenuInfoProviderState();
+}
+
+class MenuInfoProviderState extends State<MenuInfoProvider> {
```
HrX03, line 97: Is there any reason to have this state public? Refrain from using public state as you could access it using GlobalKeys (and having people mess with it)

lib/src/controls/flyouts/content_manager.dart
```diff
@@ -0,0 +1,124 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(builder: (context, constraints) {
+      return widget.builder(
+        context,
+        constraints,
+        _menus.values,
+        _menus.keys,
+      );
+    });
```
HrX03, line 122: nit: trailing comma

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+      if (isRightPreferred) return FlyoutPlacementMode.topRight;
+    }
+
+    return configuration.preferredMode;
+  }
+}
+
+extension on double {
+  /// If negative, 0 is returned
+  double _ensurePositive() {
```
HrX03, line 238: .clamp(0, double.infinity)?

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+        assert(
+          horizontal != null && horizontal
+              ? preferredMode == FlyoutPlacementMode.left ||
+                  preferredMode == FlyoutPlacementMode.right
+              : true,
+          'If the mode horizontal, preferredMode must either be left or right',
+        ),
+        assert(autoAvailableSpace == null || !autoAvailableSpace.isNegative),
+        horizontal = horizontal ??
+            [FlyoutPlacementMode.left, FlyoutPlacementMode.right]
```
HrX03, line 273: Prefer having this as a getter so you can guarantee const-ness of this constructor

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+  /// Whether the flyout should be displayed horizontally
+  ///
+  /// If true, [preferredMode] must be either .left or .right
+  final bool horizontal;
+
+  /// The preferred mode
+  final FlyoutPlacementMode preferredMode;
+
+  /// The configuration for flyout auto mode
+  FlyoutAutoConfiguration({
```
HrX03, line 259: Consider making this constructor const (by also using this.horizontal)

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+                .contains(preferredMode);
+}
+
+/// A delegate for computing the layout of a flyout to be displayed according to
+/// a target specified in the global coordinate system.
+class _FlyoutPositionDelegate extends SingleChildLayoutDelegate {
+  /// Creates a delegate for computing the layout of a flyout.
+  ///
+  /// The arguments must not be null.
+  _FlyoutPositionDelegate({
```
HrX03, line 283: Prefer const constructors

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+
+    _open = true;
+    notifyListeners();
+
+    final flyoutKey = GlobalKey();
+
+    final result = await navigator.push<T>(PageRouteBuilder<T>(
+      opaque: false,
+      transitionDuration: transitionDuration,
+      pageBuilder: (context, animation, secondary) {
```
HrX03, line 576: Make this a separate method for better readability

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+        targetBox.localToGlobal(Offset.zero, ancestor: navigatorBox) +
+            Offset(0, targetSize.height);
+    final targetRect =
+        targetBox.localToGlobal(Offset.zero, ancestor: navigatorBox) &
+            targetSize;
+
+    _open = true;
+    notifyListeners();
+
+    final flyoutKey = GlobalKey();
```
HrX03, line 571: I don't know if having a global key live in this scope is safe, can you confirm?

lib/src/controls/flyouts/flyout.dart
```diff
@@ -0,0 +1,779 @@
...
+
+    _open = false;
+    notifyListeners();
+
+    return result;
+  }
+}
+
+class _DismissAction extends DismissAction {
+  _DismissAction(this.onDismiss);
```
HrX03, line 729: If possible make const

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+/// ![MenuFlyout](https://docs.microsoft.com/en-us/windows/apps/design/controls/images/contextmenu_rs2_icons.png)
+///
+/// See also:
+///
+///  * [FlyoutController], which displays a flyout to the given target
+///  * <https://learn.microsoft.com/en-us/windows/apps/design/controls/menus>
+class MenuFlyout extends StatefulWidget {
+  /// Creates a menu flyout.
+  const MenuFlyout({
+    Key? key,
```
HrX03, line 19: use super.key instead

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    this.color,
+    this.shape,
+    this.shadowColor = Colors.black,
+    this.elevation = 8.0,
+    this.constraints,
+    this.padding = const EdgeInsets.only(top: 8.0),
+  }) : super(key: key);
+
+  /// {@template fluent_ui.flyouts.menu.items}
+  /// The colletion used to generate the content of the menu.
```
HrX03, line 30: Maybe be more esplicative on how this is used? For example could specify this is a list and also that items are arranged in a vertical list

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    super.initState();
+    generateKeys();
+  }
+
+  // @override
+  // void didUpdateWidget(covariant MenuFlyout oldWidget) {
+  //   super.didUpdateWidget(oldWidget);
+
+  //   if (widget.items != oldWidget.items) generateKeys();
+  // }
```
HrX03, line 95: Was this mistakenly pushed?

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+  /// clutter, and prevent users from focusing on the important menu items.
+  ///
+  /// ![](https://learn.microsoft.com/en-us/windows/apps/design/controls/images/contextmenu_rs2_icons.png)
+  ///
+  /// Usually an [Icon] widget,
+  final Widget? leading;
+  final Widget text;
+  final Widget? trailing;
+  final VoidCallback? onPressed;
+  final bool selected;
```
HrX03, line 247: Why not comment these too?

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+
+  bool _useIconPlaceholder = false;
```
HrX03, line 249: Where is this used? Consider removing this to allow this class to be const

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+  Widget build(BuildContext context) {
+    final size = ContentSizeInfo.of(context).size;
+    return Container(
+      width: size.isEmpty ? null : size.width,
+      padding: MenuFlyout.itemsPadding,
+      child: FlyoutListTile(
+        selected: selected,
+        showSelectedIndicator: false,
+        icon: leading ??
+            () {
```
HrX03, line 261: Avoid this kind of lambda calls for logic, either make a new function for this, compute this before returning or use a ternary operator. In this case i'd suggest removing this logic entirely and just returning null

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    super.key,
+    super.leading,
+    required super.text,
+    super.trailing = const Icon(FluentIcons.chevron_right),
+    required this.items,
+    this.showBehavior = SubItemShowBehavior.hover,
+    this.showHoverDelay = const Duration(milliseconds: 600),
+  }) : super(onPressed: null);
+
+  GlobalKey<__MenuFlyoutSubItemState>? _key;
```
HrX03, line 336: Could you explain this?

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    Key? key,
+    required this.item,
+    required this.items,
+  }) : super(key: key);
+
+  @override
+  State<_MenuFlyoutSubItem> createState() => __MenuFlyoutSubItemState();
+}
+
+class __MenuFlyoutSubItemState extends State<_MenuFlyoutSubItem>
```
HrX03, line 373: Avoid double leading underscores, just use one

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+  void dispose() {
+    transitionController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final menuInfo = MenuInfoProvider.of(context);
+
+    transitionController.duration = menuInfo.widget.transitionDuration;
```
HrX03, line 394: this feels quite hacky, please rethink this logic to use a better solution, for example you could pass this parameter from the constructor itself and just use didUpdateWidget to update the duration

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+
+    return MouseRegion(
+      onEnter: (event) {
+        showTimer = Timer(widget.item.showHoverDelay, () {
+          show(menuInfo);
+        });
+      },
+      onExit: (event) {
+        if (showTimer != null && showTimer!.isActive) {
+          showTimer!.cancel();
```
HrX03, line 404: You can just do showTimer?.cancel(), it's always safe to call this method

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    transitionController.forward();
+    setState(() {});
+  }
+
+  /// Closes this menu and its children
+  Future<void> close(MenuInfoProviderState menuInfo) async {
+    await Future.wait([
+      if (menuKey.currentState != null)
+        ...menuKey.currentState!.keys
+            .whereType<GlobalKey<__MenuFlyoutSubItemState>>()
```
HrX03, line 461: I suggest using a provider that lives on top of the menu to have a cleaner architecture, this can and will break very easily. Use something like a `MenuSubitemRegistry` to keep track of current open subitems

lib/src/controls/flyouts/menu.dart
```diff
@@ -0,0 +1,532 @@
...
+    if (mounted) setState(() {});
+  }
+}
+
+class _SubItemPositionDelegate extends SingleChildLayoutDelegate {
+  final Rect parentRect;
+  final Size parentSize;
+  final double margin;
+
+  _SubItemPositionDelegate({
```
HrX03, line 477: Make the constructor const

lib/src/controls/inputs/dropdown_button.dart
```diff
@@ -203,15 +198,62 @@ class _DropDownButtonState extends State<DropDownButton> {
...
+
+  void showFlyout() async {
+    widget.onOpen?.call();
+    await flyoutController.showFlyout(
+      placementMode: FlyoutPlacementMode.auto,
+      autoModeConfiguration: FlyoutAutoConfiguration(
+        preferredMode: widget.placement,
+      ),
+      additionalOffset: widget.verticalOffset,
+      transitionBuilder: (context, animation, placement, flyout) {
```
HrX03, line 212: Move this onto its own method?

lib/src/controls/navigation/navigation_view/pane_items.dart
```diff
@@ -664,6 +663,13 @@ class __PaneItemExpanderState extends State<_PaneItemExpander>
     if (_open) {
       controller.value = 1;
     }
+
+    flyoutController.addListener(() {
```
HrX03, line 667: Be wary! This method will be called each time one of the inherited widgets you depend on in the build method has changed (and signals it has indeed) so you could end up adding listeners multiple times. Add some sort of check or move this call

lib/src/controls/navigation/navigation_view/pane_items.dart
```diff
@@ -679,20 +685,73 @@ class __PaneItemExpanderState extends State<_PaneItemExpander>
...
+                      margin: const EdgeInsetsDirectional.only(bottom: 4.0),
+                      child: DefaultTextStyle(
+                        style: navigationTheme.itemHeaderTextStyle ??
+                            const TextStyle(),
+                        softWrap: false,
+                        maxLines: 1,
+                        overflow: TextOverflow.fade,
+                        // textAlign: view.displayMode == PaneDisplayMode.top
+                        //     ? TextAlign.center
+                        //     : TextAlign.left,
```
HrX03, line 736: Remove commented code if not needed

lib/src/controls/navigation/navigation_view/pane_items.dart
```diff
@@ -679,20 +685,73 @@ class __PaneItemExpanderState extends State<_PaneItemExpander>
...
+                  return MenuFlyoutItemBuilder(builder: (context) {
+                    return Container(
+                      padding: const EdgeInsets.symmetric(
+                        horizontal: 10.0,
+                        vertical: 8.0,
+                      ),
+                      margin: const EdgeInsetsDirectional.only(bottom: 4.0),
+                      child: DefaultTextStyle(
+                        style: navigationTheme.itemHeaderTextStyle ??
+                            const TextStyle(),
```
HrX03, line 730: Is this actually needed? As in you're not allowed to pass null to it

lib/src/controls/navigation/navigation_view/pane_items.dart
```diff
@@ -679,20 +685,73 @@ class __PaneItemExpanderState extends State<_PaneItemExpander>
...
+
+        flyoutController.showFlyout(
+          placementMode: displayMode == PaneDisplayMode.compact
+              ? FlyoutPlacementMode.right
+              : FlyoutPlacementMode.bottomCenter,
+          builder: (context) {
+            return MenuFlyout(
+              items: widget.items.map<MenuFlyoutItemBase>((item) {
+                if (item is PaneItem) {
+                  return _PaneItemExpanderMenuItem(
```
HrX03, line 710: Consider adding a method like `getMenuItem` to PaneItem so you can remove this whole switch/case statement

example/lib/screens/popups/flyout.dart
```diff
@@ -0,0 +1,390 @@
...
+                        value: mode,
+                        child: Text(mode.name.uppercaseFirst()),
+                      );
+                    }).toList(),
+                    value: placementMode,
+                    onChanged: (mode) {
+                      if (mode != null) setState(() => placementMode = mode);
+                    },
+                  ),
+                ]),
```
HrX03, line 96: nit: Trailing comma

example/lib/screens/popups/flyout.dart
```diff
@@ -0,0 +1,390 @@
...
+                    ),
+                  );
+                  // return MenuFlyout(
+                  //   items: [
+                  //     MenuFlyoutItem(
+                  //       text: const Text('TEST'),
+                  //       onPressed: () {},
+                  //     ),
+                  //   ],
+                  // );
```
HrX03, line 361: Remove commented code if not needed

HrX03 (commented):
That's most of the stuff i found, some things could be expanded and explained better if needed, you know where to find me in case :)

lib/src/controls/flyouts/content_manager.dart
bdlukaa: Never found any other way to do this

lib/src/controls/flyouts/content_manager.dart
bdlukaa: To check the size of your flyout, you can use `ContentSizeInfo.of(context).size`

lib/src/controls/flyouts/flyout.dart
bdlukaa: smart

lib/src/controls/flyouts/flyout.dart
bdlukaa: yeah, it is safe because we push a new route.

lib/src/controls/flyouts/menu.dart
bdlukaa: I believe it's off the scope of this PR (and I'm lazy)

lib/src/controls/flyouts/menu.dart
bdlukaa: Allows the following ('Rename', 'Select').

image

> Consider removing this to allow this class to be const

It's not worth it because of [onPressed]

lib/src/controls/flyouts/menu.dart
bdlukaa: It is the key of `_MenuFlyoutSubItem`, built in the `build` method. It can not use the parent `key` widget because it's already used by `MenuFlyout` to build the widget. It is assigned by `MenuFlyout` with a key generated by `generateKeys()`. It is used to close the child menu when its parent is closed

lib/src/controls/flyouts/menu.dart
bdlukaa: If we reach a null pointer error, something is wrong. It's good to keep this to know where the issue comes, if any, in the future

lib/src/controls/inputs/dropdown_button.dart
bdlukaa: Don't see why

lib/src/controls/navigation/navigation_view/pane_items.dart
bdlukaa: Yeah, `DefaultTextStyle.style` doesn't allow a null style

lib/src/controls/flyouts/content_manager.dart
bdlukaa: By calling `MenuInfoProvider.of(context)`, it returns this state class. It can close and add sub-flyouts to the current flyouts

lib/src/controls/flyouts/menu.dart
bdlukaa: >  In this case i'd suggest removing this logic entirely and just returning null

`Icon(null)` makes the box have the same size as if it had an icon data. It's used if there is a placeholder

lib/src/controls/flyouts/content_manager.dart
HrX03: I think it would be safer to provide a view of this class instead of the state itself, it's safer

PR #730, opened 2023-02-09
Tabview updates

### LINKED ISSUE #648: 🐛 TextBox focus behavior inside Tab
**Describe the bug**
Textboxes can be focused by clicking on them, and this focus is removed by clicking anywhere else in the application (usually). However, when a textbox is put inside a Tab, this focus removal is not possible anymore. 

**To Reproduce**
1. Copy [the following gist](https://gist.github.com/Daanchaam/32e6c1fece3a69570fffef700d3c622f) into a new flutter application with fluent_ui as a dependency
2. Run the application, click the first TextBox
3. Click outside the TextBox, note the focus is not being removed 

**Expected behavior**
The focus should be removed from the TextBox, as it does when not wrapped inside a TabView

**Additional context**
I have tried options like [this one](https://stackoverflow.com/questions/53481261/how-to-unfocus-textfield-that-has-custom-focusnode), but to no avail yet. It definitely feels like this is not intended behavior.

### LINKED ISSUE #728: tabview support macos shortcut
tabview support macos shortcuts
- command + w
- command + t
- ...

bdlukaa:
Fixes #648
Fixes #728

## Pre-launch Checklist


lib/src/controls/navigation/tab_view.dart
```diff
@@ -595,17 +598,33 @@ class _TabViewState extends State<TabView> {
         close(widget.currentIndex);
       }
 
+      // For more info, refer to [SingleActivator] docs
+      var ctrl = true;
+      var meta = false;
+      if (!kIsWeb && defaultTargetPlatform == TargetPlatform.iOS) {
```
h3x4d3c1m4l, line 604: Shouldn't this be both iOS and macOS?

bdlukaa: nice catch

PR #753, opened 2023-03-04
add build workflows

bdlukaa:
## Pre-launch Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
         run: flutter pub get
         working-directory: ./example
       - name: Build for Windows
-        run: flutter build windows example/lib/main.dart --verbose
+        run: flutter build windows --verbose
         working-directory: ./example
 
   # build_linux:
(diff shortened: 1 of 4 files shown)


h3x4d3c1m4l:
lmk if you would like any help with macOS!

bdlukaa:
I don't think it makes sense to have a build workflow for platforms other than windows

h3x4d3c1m4l:
Well looking from "Microsoft perspective", it probably doesn't cause Fluent UI is designed for Windows. However, due to Flutter's nature being a multiplatform UI framework, it could make sense. There is nothing in `fluent_ui` that stops me from running my `fluent_ui` based apps on macOS. It even looks pretty nice too 😄 

On the other hand there are pretty good reasons you should indeed ignore other platforms:
- As of now you still need macOS to compile for macOS, and macOS minutes on GH Actions are expensive
-

bdlukaa:
I ended up adding the build workflow for it, just for precaution.

> As of now you still need macOS to compile for macOS, and macOS minutes on GH Actions are expensive

They aren't charged on public repos

PR #771, opened 2023-03-15
Add NumberBox Widget

WinXaito:
## Pre-launch Checklist

Need to do:

- [ ] Update value with mouse scroll 
- [ ] Update value with `PageUp` / `PageDown` shortcut
- [ ] Add documentation
- [ ] Add `disabled` options
- [ ] Theme ?? 

Demo: 

numberbox

### DIFF
--- a/lib/src/controls/form/number_box.dart
+++ b/lib/src/controls/form/number_box.dart
@@ -1,0 +1,291 @@
+import 'package:fluent_ui/fluent_ui.dart';
+import 'package:flutter/services.dart';
+
+enum SpinButtonPlacementMode {
+  inline,
+  compact,
+}
+
+class NumberBox extends StatefulWidget {
+  final int? value;
+  final ValueChanged<int?>? onChanged;
+  final String placeholderText;
+  final SpinButtonPlacementMode mode;
+  final FocusNode? focusNode;
+
+  final bool clearButton;
+
+  final int smallChange;
+  final int largeChange;
+
+  const NumberBox({
+    super.key,
+    required this.value,
+    required this.onChanged,
+    this.placeholderText = '',
+    this.focusNode,
+    this.mode = SpinButtonPlacementMode.compact,
+    this.clearButton = true,
+    this.smallChange = 1,
+    this.largeChange = 10,
+  });
+
+  @override
+  State<StatefulWidget> createState() => _NumberBoxState();
+}
+
+class _NumberBoxState extends State<NumberBox> {
+  FocusNode? _internalNode;
+
+  FocusNode? get focusNode => widget.focusNode ?? _internalNode;
+
+  OverlayEntry? _entry;
+
+  bool _hasPrimaryFocus = false;
+
+  final controller = TextEditingController();
+
+  bool clearBox = false;
+
+  final LayerLink _layerLink = LayerLink();
+  final GlobalKey _textBoxKey = GlobalKey(
+    debugLabel: "NumberBox's TextBox Key",
+  );
+
+  // Only used if needed to create _internalNode.
+  FocusNode _createFocusNode() {
+    return FocusNode(debugLabel: '${widget.runtimeType}');
+  }
+
+  @override
+  void initState() {
+    if (widget.focusNode == null) {
+      _internalNode ??= _createFocusNode();
+    }
+    focusNode!.addListener(_handleFocusChanged);
+
+    controller.text = widget.value?.toString() ?? '';
+    super.initState();
+  }
+
+  @override
+  void dispose() {
+    focusNode!.removeListener(_handleFocusChanged);
+    _internalNode?.dispose();
+    super.dispose();
+  }
+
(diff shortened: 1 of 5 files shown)


lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+      enabled: widget.onChanged != null,
+      suffix:
+          textFieldSuffix.isNotEmpty ? Row(children: textFieldSuffix) : null,
+    );
+
+    return CompositedTransformTarget(
+      link: _layerLink,
+      child: Focus(
+        onKeyEvent: (node, event) {
+          if (event is! KeyDownEvent) {
```
bdlukaa, line 271: `KeyRepeatEvent` should be allowed. This allows to keep the key pressed and keep incrementing

lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+            return KeyEventResult.ignored;
+          }
+
+          if (event.logicalKey == LogicalKeyboardKey.pageUp) {
+            _incrementLarge();
+          } else if (event.logicalKey == LogicalKeyboardKey.pageDown) {
+            _decrementLarge();
+          }
```
bdlukaa, line 279: `.arrowUp` and `.arrowDown` should be added in the condition. My laptop doesn't have a `pageUp` button, so it doesn't work there, but the arrow buttons work. `AutoSuggestBox` uses them to manipulate focus

lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+
+          return KeyEventResult.ignored;
```
bdlukaa, line 281: Not sure, but shouldn't it be `return KeyEventResult.handled;` when the conditions are met?

lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+    this.clearButton = true,
+    this.smallChange = 1,
+    this.largeChange = 10,
+  });
+
+  @override
+  State<StatefulWidget> createState() => _NumberBoxState();
+}
+
+class _NumberBoxState extends State<NumberBox> {
```
bdlukaa, line 100: maybe exposing the state class as public and the increment functions, so that it is possible to manipulate it with a `GlobalKey<NumberBoxState>`?

lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+    required this.onChanged,
+    this.focusNode,
+    this.mode = SpinButtonPlacementMode.compact,
+    this.clearButton = true,
+    this.smallChange = 1,
+    this.largeChange = 10,
+  });
+
+  @override
+  State<StatefulWidget> createState() => _NumberBoxState();
```
bdlukaa, line 97:
```suggestion
  State<NumberBox> createState() => _NumberBoxState();
```

lib/src/controls/form/number_box.dart
```diff
@@ -0,0 +1,412 @@
...
+    }
+  }
+
+  void _dismissOverlay() {
+    _entry?.remove();
+    _entry = null;
+  }
+
+  @override
+  Widget build(BuildContext context) {
```
bdlukaa, line 229:
```suggestion
  Widget build(BuildContext context) {
    assert(debugCheckHasFluentTheme(context));
```

bdlukaa (changes requested):
This looks awesome, but a few things are missing.

[Text formatting](https://learn.microsoft.com/en-us/windows/apps/design/controls/number-box#formatting-input). Just exposing `TextBox.inputFormatters` and adding an example should do the job.

`TextBox.keyboardType` should be number.

Exposing some other `TextBox` properties, such as `placeholder`, `autoFocus`, cursor-related properties, and other that may be relevant.

lib/src/controls/form/number_box.dart
WinXaito: Nice catch !

lib/src/controls/form/number_box.dart
WinXaito: Absolutely. I think the most natural is to have page up/down for large increments and arrow up/down for small ones

lib/src/controls/form/number_box.dart
WinXaito: I'm sorry but I don't known how to do that. Feel free to add a commit for doing that ;)

ricardoboss: Just remove the _ (underscore) from the state class name

WinXaito: Oh I see, that was the globalKey that I didn’t understand. But I see now. I will do that tomorrow

bdlukaa:
Sounds ok to me!

bdlukaa (approved):

PR #772, opened 2023-03-15
Fix tab view wheel scroll

WinXaito:
## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/navigation/tab_view.dart
+++ b/lib/src/controls/navigation/tab_view.dart
@@ -73,7 +73,7 @@ class TabView extends StatefulWidget {
     this.header,
     this.footer,
     this.closeDelayDuration = const Duration(milliseconds: 400),
-    @Deprecated('this value is no longer used and will be removed in the next major release.')
+    @Deprecated('This property is no longer used and will be removed in the next major release.')
         this.wheelScroll = false,
   }) : super(key: key);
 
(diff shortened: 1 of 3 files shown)


bdlukaa:
The `wheelScroll` property can then be removed, right?

WinXaito:
> The `wheelScroll` property can then be removed, right?

Note necessary, if a user want to keep the vertical scroll over all page he can. 

But I think we must put the default value to true. 

What do you think about that ? And if you really want to remove it, I think it’s preferable to set it to true by default and mark as deprecated to avoid a breaking change.

bdlukaa:
> And if you really want to remove it, I think it’s preferable to set it to true by default and mark as deprecated to avoid a breaking change and remove it in v5.0.0

Sounds fine to me! I checked on native and it stops the vertical scrolling, so I think we should do the same.

lib/src/controls/navigation/tab_view.dart
```diff
@@ -74,6 +73,8 @@ class TabView extends StatefulWidget {
     this.header,
     this.footer,
     this.closeDelayDuration = const Duration(milliseconds: 400),
+    @Deprecated('this value is no longer used and will be removed in the next major release.')
```
bdlukaa, line 76:
```suggestion
    @Deprecated('This property is no longer used and will be removed in the next major release.')
```

bdlukaa (changes requested):
Tests are failing because we make use of the property on the example app

WinXaito:
👍 that must be good now

bdlukaa (approved):

PR #789, opened 2023-03-24
Let numberbox to accept double value

WinXaito:
## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/form/number_box.dart
+++ b/lib/src/controls/form/number_box.dart
@@ -49,10 +49,10 @@ enum SpinButtonPlacementMode {
 /// See also:
 ///
 ///  * https://learn.microsoft.com/en-us/windows/apps/design/controls/number-box
-class NumberBox extends StatefulWidget {
+class NumberBox<T extends num> extends StatefulWidget {
   /// The value of the number box. When this value is null, the number box field
   /// is empty.
-  final int? value;
+  final T? value;
 
   /// Called when the value of the number box change.
   /// The callback is fired only if the user click on a button or the focus is
@@ -61,7 +61,7 @@ class NumberBox extends StatefulWidget {
   /// If the [onChanged] callback is null then the number box widget will
   /// be disabled, i.e. its buttons will be displayed in grey and it will not
   /// respond to input.
-  final ValueChanged<int?>? onChanged;
+  final ValueChanged<T?>? onChanged;
 
   /// {@macro flutter.widgets.Focus.focusNode}
   final FocusNode? focusNode;
@@ -75,12 +75,12 @@ class NumberBox extends StatefulWidget {
 
   /// The value that is incremented or decremented when the user click on the
   /// buttons or when he scroll on the number box.
-  final int smallChange;
+  final num smallChange;
 
   /// The value that is incremented when the user click on the shortcut
   /// [LogicalKeyboardKey.pageUp] and decremented when the user lick on the
   /// shortcut [LogicalKeyboardKey.pageDown].
-  final int largeChange;
+  final num largeChange;
 
   /// A widget displayed at the start of the text box
   ///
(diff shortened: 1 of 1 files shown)


lib/src/controls/form/number_box.dart
```diff
@@ -75,12 +78,30 @@ class NumberBox extends StatefulWidget {
...
-  final int largeChange;
+  final num largeChange;
+
+  /// The precision indicates the number of digits that's accepted for double
+  /// value.
+  final int precision;
+
+  /// The minimum value allowed. If the user input a value below than min,
+  /// the value is replaced by min.
+  /// If min is null, there is lowest limit.
```
bdlukaa, line 95:
```suggestion
  /// If min is null, there is no lowest limit.
```

lib/src/controls/form/number_box.dart
```diff
@@ -75,12 +78,30 @@ class NumberBox extends StatefulWidget {
...
+  final num? min;
+
+  /// The maximum value allowed. If the user input a value greater than max,
+  /// the value is replaced by max.
+  /// If max is null, there is no upper limit.
+  final num? max;
+
+  /// When true, if something else than a number is specified, the content of
+  /// the text field is interpreted as an expressions when the focus is lost.
+  final bool allowExpressions;
```
bdlukaa, line 105:
```suggestion
  /// When true, if something else than a number is specified, the content of
  /// the text box is interpreted as a math expression when the focus is lost.
  /// 
  /// See also:
  ///
  ///   * <https://pub.dev/packages/math_expressions>
  final bool allowExpressions;
```

bdlukaa (commented):
Maybe make some methods available, so that it's possible to change the numberbox value programatically using a global key?

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## [next]
 
 - Add `NumberBox` widget. ([#560](https://github.com/bdlukaa/fluent_ui/issues/560) [#771](https://github.com/bdlukaa/fluent_ui/pull/771))
+- Add `min`/`max` properties to the `NumberBox` widget. ([#789](https://github.com/bdlukaa/fluent_ui/pull/789))
+- Allow `NumberBox` Widget to accept int and double values and mathematical expressions. ([#789](https://github.com/bdlukaa/fluent_ui/pull/789))
```
bdlukaa, line 5: I am not sure we need to add these to the changelog since the version hasn't been published. Wdyt?

lib/src/controls/form/number_box.dart
```diff
@@ -168,7 +198,7 @@ class NumberBoxState extends State<NumberBox> {
 
   bool _hasPrimaryFocus = false;
 
-  late int? previousValidValue = widget.value;
+  late num? previousValidValue = widget.value;
 
   final controller = TextEditingController();
```
bdlukaa, line 203: It looks like this controller isn't being disposed

lib/src/controls/form/number_box.dart
```diff
@@ -384,54 +419,88 @@ class NumberBoxState extends State<NumberBox> {
...
+    final value = (num.tryParse(controller.text) ?? widget.value ?? 0) -
         widget.largeChange;
-    _updateController(value);
-    _updateValue();
+    updateController(value);
+    updateValue();
   }
 
-  void _updateController(int value) {
+  void updateController(num value) {
```
bdlukaa, line 453: I think this method can be private. I don't see any public usages for this

lib/src/controls/form/number_box.dart
```diff
@@ -384,54 +419,88 @@ class NumberBoxState extends State<NumberBox> {
...
     }
     previousValidValue = value;
 
     if (widget.onChanged != null) {
-      widget.onChanged!(value);
+      widget.onChanged!(value as T?);
+    }
+  }
+
+  String? format(num? value) {
```
bdlukaa, line 496: I think this can also be private

CHANGELOG.md
WinXaito: Yep, I just added the reference to this PR in the Changelog

bdlukaa (approved):

PR #791, opened 2023-03-24
Added min and max properties to numberbox

ricardoboss:
I added `min` and `max` properties to the new `NumberBox` widget. These allow users to constrain the value that can be entered using a number.

## Pre-launch Checklist

I haven't updated the CHANGELOG since the `NumberBox` widget has its own point already.

### DIFF
--- a/lib/src/controls/form/number_box.dart
+++ b/lib/src/controls/form/number_box.dart
@@ -82,6 +82,14 @@ class NumberBox extends StatefulWidget {
   /// shortcut [LogicalKeyboardKey.pageDown].
   final int largeChange;
 
+  /// The maximum value that can be set. Any means to change the value will
+  /// limit the upper bound to this value. Set to [null] to remove the limit.
+  final int? max;
+
+  /// The minimum value that can be set. Any means to change the value will
+  /// limit the lower bound to this value. Set to [null] to remove the limit.
+  final int? min;
+
   /// A widget displayed at the start of the text box
   ///
   /// Usually an [IconButton] or [Icon]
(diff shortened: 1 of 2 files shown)


lib/src/controls/form/number_box.dart
```diff
@@ -415,7 +425,21 @@ class NumberBoxState extends State<NumberBox> {
...
+    if (widget.min != null && widget.min! > value) {
+      return widget.min!;
+    }
+
+    if (widget.max != null && widget.max! < value) {
+      return widget.max!;
+    }
+
+    return value;
+  }
```
bdlukaa, line 438: I would make use of the dart's math lib to shortcut this code

```suggestion
  int _constrainValue(int value) {
    return math.min(
      widget.min ?? double.infinity,
      math.max(
        widget.max ?? double.infinity,
        value,
      ),
    );
  }
```

ricardoboss: Not sure that's more readable, but shorter nonetheless.

bdlukaa (commented):
The `didUpdateWidget` should be added and with it a check to see if the `max` and `min` properties have changed. If so, constrain to the bounds

WinXaito:
Hm, I already do that here: with generic and math expressions https://github.com/bdlukaa/fluent_ui/pull/789#pullrequestreview-1357366578

sorry if the title of the PR was incomplete..

lib/src/controls/form/number_box.dart
henry2man: @bdlukaa I agree with @ricardoboss, his code looks simpler and more readable 😅

ricardoboss:
@WinXaito ah, well then this PR is needless.

bdlukaa:
I will close it then :)

PR #793, opened 2023-03-28
Load the example app with a bigger size for have a open navigation pane and do not center the window

WinXaito:
I find it really annoying that the default Window size as a navigation pane is compact mode.

And each time a do a hot restart, I must change window size.

And at each hot restart, the window is replaced on the center. If we do not enable the center, with an hot restart, the window keep it's position.

---------------

@bdlukaa I know that you work on a laptop. Do you think these settings are good for you ? (For my part, I work on a 39" curved screen, so...)

--------------

Before:
image

After:
image

## Pre-launch Checklist

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -51,9 +51,8 @@ void main() async {
         TitleBarStyle.hidden,
         windowButtonVisibility: false,
       );
-      await windowManager.setSize(const Size(755, 545));
+      await windowManager.setSize(const Size(1050, 545));
       await windowManager.setMinimumSize(const Size(350, 600));
-      await windowManager.center();
       await windowManager.show();
       await windowManager.setPreventClose(true);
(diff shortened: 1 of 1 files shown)


example/lib/main.dart
```diff
@@ -51,9 +51,8 @@ void main() async {
         TitleBarStyle.hidden,
         windowButtonVisibility: false,
       );
-      await windowManager.setSize(const Size(755, 545));
```
bdlukaa, line 54: Maybe we don't need to make use of `.setSize`. We can set the window size from the windows runner. That is a better solution, that works at runtime and hot restart isn't going to break it

WinXaito:
Okay, I just removed the `setSize` (currently in windows\runner\main.cpp, the size is defined to 1280. That sounds good to me).

And I changed the min size:

Current:
image

New min size:
image

bdlukaa (approved):

PR #795, opened 2023-03-29
Add PasswordBox widget

WinXaito:
This PR add a new widget: `PasswordBox` : https://learn.microsoft.com/en-us/windows/apps/design/controls/password-box

This looks like that:

passwordbox

There is four modes:

- `peek` : default mode, when the user got the focus (for the first time !) and the text field is not empty, a reveal button is added. If the focus is lost and re-got, the button is not visible (same operation than Microsoft WinUI). The button is visible again only if the text field is cleared.
- `peekAlways`: (not a Microsoft standard mode). The button is visible when the text-field is not empty and doesn't look if it the first focus or not).
- `visible` : the password is always visible (useful for a programmatically change)
- `hidden`: the password is always hidden (useful for a programmatically change)

-----------------

This PR also add parameters `onTapDown` and `onTapUp` on all buttons.

## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/form/password_box.dart
+++ b/lib/src/controls/form/password_box.dart
@@ -178,6 +178,17 @@ class _PasswordBoxState extends State<PasswordBox> {
     super.dispose();
   }
 
+  @override
+  void didUpdateWidget(covariant PasswordBox oldWidget) {
+    if (oldWidget.revealMode == PasswordRevealMode.peekAlways &&
+        widget.revealMode == PasswordRevealMode.peek) {
+      // if the mode change, we consider that the first focus is gone.
+      focusCanPeek = false;
+    }
+
+    super.didUpdateWidget(oldWidget);
+  }
+
   void _handleTextChange() {
     if (controller.text.isEmpty) {
       // If the text box is empty, then we ignore if the focus has been
--- a/lib/src/controls/form/password_box.dart
+++ b/lib/src/controls/form/password_box.dart
@@ -119,6 +119,7 @@ class PasswordBox extends StatefulWidget {
   /// {@macro flutter.widgets.editableText.obscuringCharacter}
   final String obscuringCharacter;
 
+  /// Creates a password box
   const PasswordBox({
     super.key,
     this.controller,
@@ -150,11 +151,11 @@ class _PasswordBoxState extends State<PasswordBox> {
   bool focusCanPeek = true;
   bool textCanPeek = false;
 
-  late final TextEditingController controller =
-      widget.controller ?? TextEditingController();
+  TextEditingController? _internalController;
+  TextEditingController? get controller =>
+      widget.controller ?? _internalController;
 
   FocusNode? _internalNode;
-
   FocusNode? get focusNode => widget.focusNode ?? _internalNode;
 
   bool get _isVisible =>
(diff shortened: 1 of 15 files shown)


WinXaito:
There is one think, I didn't found this icon: 
image

So I used this one for the moment:

image

bdlukaa (commented):
LGTM overall

lib/src/controls/form/password_box.dart
```diff
@@ -0,0 +1,258 @@
...
+      _internalNode ??= _createFocusNode();
+    }
+    controller.addListener(_handleTextChange);
+    focusNode!.addListener(_handleFocusChange);
+    super.initState();
+  }
+
+  @override
+  void dispose() {
+    controller.dispose();
```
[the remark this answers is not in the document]
WinXaito: So, does this must be the case for the FocusNode too ? (If this is the case, I will must update the NumberBox too)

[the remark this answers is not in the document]
bdlukaa: I think so

lib/src/controls/form/password_box.dart
```diff
@@ -0,0 +1,284 @@
...
+          textCanPeek);
+
+  @override
+  void initState() {
+    if (widget.focusNode == null) {
+      _internalNode = FocusNode(debugLabel: '${widget.runtimeType}');
+    }
+    if (widget.controller == null) {
+      _internalController = TextEditingController();
+    }
```
bdlukaa, line 180: We also must do this check on `didUpdateWidget`. It is a common case when debugging to attach a controller/node after the widget was already built (hot reload)

lib/src/controls/form/password_box.dart
```diff
@@ -0,0 +1,284 @@
...
+
+/// A fluent design input form for password.
+///
+/// A PasswordBox lets the user enter a password. There is multiple mode to
+/// indicate if the password must be hidden, visible or if a reveal button is
+/// shown.
+///
+/// See also:
+///
+///  * https://learn.microsoft.com/en-us/windows/apps/design/controls/password-box
```
bdlukaa, line 37:
```suggestion
/// A fluent design input form for password.
///
/// A password box is a text input box that conceals the characters typed into
/// it for the purpose of privacy. A password box looks like a text box, except 
/// that it renders placeholder characters in place of the text that has been 
/// entered. You can configure the placeholder character.
///
/// ![PasswordBox](https://learn.microsoft.com/en-us/windows/apps/design/controls/images/passwordbox-focus-typing.png)
///
/// See also:
///
///  * <https://learn.microsoft.com/en-us/windows/apps/design/controls/password-box>
///  * [PasswordRevealMode], the different modes that the password box can have
///  * [TextBox], the underlaying widget that renders the text box
```

lib/src/controls/form/password_box.dart
```diff
@@ -0,0 +1,284 @@
...
+class PasswordBox extends StatefulWidget {
+  /// Controls the text being edited.
+  ///
+  /// If null, this widget will create its own [TextEditingController].
+  final TextEditingController? controller;
+
+  /// Disables the text field when false.
+  ///
+  /// Text fields in disabled states have a light grey background and don't
+  /// respond to touch events including the [prefix], [suffix].
```
bdlukaa, line 47: We don't have prefix nor suffix

```suggestion
  /// respond to touch events.
```

bdlukaa (approved):

PR #815, opened 2023-04-24
Solve library not support intl v 0.18 

the-best-is-best:


bdlukaa:
Nice catch! Please do not update the version of the package

the-best-is-best:
> Nice catch! Please do not update the version of the `fluent_ui` package

ok i return version of fluent_ui and remove all changed in changelog file

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: fluent_ui
 description: Implements Windows UI in Flutter. Based on the official documentation
-version: 4.5.0
+version: 4.5.1
```
bdlukaa, line 3:
```suggestion
version: 4.5.0
```

this should be maintained as it is

the-best-is-best: Ok sorry i solved it

WinXaito (approved):
Lgtm

PR #840, opened 2023-05-13
Support flutter 3.10.x

the-best-is-best:
Minimum dart 3 is required


the-best-is-best:
Please accept this changes need minimum Dart3 and flutter 3.10 because the changes in text field will not support dart under 3

bdlukaa (approved):

PR #855, opened 2023-06-06
Proper implementation on localizing date and time with `intl` package

rk0cc:
Replace original mixin implementation in #445 to [my published package](https://pub.dev/packages/intl_script_recognizer) and cover all codes which required `DateFormat`.

And I also fixed the constraint of `intl` package properly.

## Pre-launch Checklist


bdlukaa:
Sorry to say, but I don't understand the issue with the current implementation and, consequently, I don't understand what problem this is fixing. Could you elaborate?

rk0cc:
I just replace the original mixin solution into a deliciated pub package that considering it may be necessary if some user implements into other themes of app for better fallback procedures. In additions, I'm also make it cover to all DateTime which require perform localization with `Locale` object that ensure the date and time will be display according to the language and country code in corresponded script language if required.

For the version constraint of `intl`, although it does not have any issue of implementation right now, it likely to break other implemented package when the dependency release beyond `1.0.0`. As a result, it is better to uses caret constraint instead of applying `<=1.0.0`.

PR #872, opened 2023-07-01
feat: [CommandBar] supports setting directions if needed.

CorvusYe:
Add direction to CommandBar to support verticality.
Sometimes, we need to place the command bar on the side. 
Just like this:
image

When we set `direction` to `Axis.vertical`, the `isCompact` will default to true, if it is not set.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## [next]
 
+- Add vertical support to `CommandBar`. ([#872](https://github.com/bdlukaa/fluent_ui/pull/872))
 - **BREAKING** Removed `SplitButtonBar` and its related widget. Use `SplitButton` or `SplitButton.toggle` instead ([#882](https://github.com/bdlukaa/fluent_ui/pull/882), [#411](https://github.com/bdlukaa/fluent_ui/issues/411))
 
 ## 4.7.0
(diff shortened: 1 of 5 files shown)


CorvusYe:
Sorry, this version can't cover all options, uch as `overflowBehavior: CommandBarOverflowBehavior.dynamicOverflow`.

If you agree with this feature, I will make another PR when all options are compatible.

Sorry for the previous mistake again.

bdlukaa:
I like the concept and would love to see this working :)

CorvusYe:
@bdlukaa 
Now it is compatible with scrolling and dynamicOverflow.
But, I have a question about whether to expand the `scrollDirection` attribute in `HorizontalScrollView`？
This makes me hesitate.

bdlukaa:
`HorizontalScrollView` is just a helper widget that makes scrolling horizontally smoother. I don't see a problem using only `SingleChildScrollView` on vertical command bars. 

I'm not sure how resolving the mouse scroll delta for a better scrolling effect works for vertical scroll views tho. iirc, Flutter is working on 2D scrolling, which should affect how horizontal scrolling behaves and we, sometime soon, may get rid of `HorizontalScrollView` at all.

CorvusYe:
> Flutter is working on 2D scrolling, which should affect how horizontal scrolling behaves and we, sometime soon, may get rid of HorizontalScrollView at all.

Does this mean that `HorizontalScrollView` may be deprecated in the future?
The performance experience of `SingleChildScrollView` in the vertical direction is still inferior to that in the horizontal direction.

Here are some effects of using vertical scrolling.
---
- When I use the mouse scroll wheel in normal mode, there is a slight jumping sensation.

https://github.com/bdlukaa/fluent_ui/assets/15630211/313c8517-a684-4923-96ae-4e2961ead3cd

- But when I hold down the `shift` key, the performance is as smooth as when it's horizontal.


https://github.com/bdlukaa/fluent_ui/assets/15630211/53fb2b0d-caea-4a36-8d84-aa8ccd2ae92b

CorvusYe:
What I don't quite understand is why [this event](https://github.com/bdlukaa/fluent_ui/blob/76bb450adf513eb2aaca6d237f4187f90ed368d3/lib/src/utils/horizontal_scroll_view.dart#L53-L64) only triggers when it reaches the top and bottom.

Do I need to make any further changes?

bdlukaa:
I tried your implementation with every example in https://bdlukaa.github.io/fluent_ui/#/surfaces/command_bar, but most of them throw a render box issue :/

CorvusYe:
In my case, the exterior has height constraints. Is it possible that this is the reason？

CorvusYe:
It seems that height constraints are still necessary I wrote examples of `vertical` according to the `horizontal` usage and following behind each one.
image

Do you have any thoughts on the hiding of vertical scrollbars?  It looks very oppressive now.

bdlukaa:
You can wrap the scroll view in a

```dart
ScrollConfiguration(
 behavior: ScrollConfiguration.of(context).copyWith(scrollbars: false),
 child: ...,
),
 ```

CorvusYe:
@bdlukaa 
Thanks for all the support you've given me. 
Not only for this PR, but also for my personal project which dependents fluent_ui.

PR #1023, opened 2024-02-11
TreeView item add disabled option: Whether the current item is disabled

imxood:
## Pre-launch Checklist

### DIFF
--- a/example/lib/screens/navigation/tree_view.dart
+++ b/example/lib/screens/navigation/tree_view.dart
@@ -122,7 +122,7 @@ TreeView(
 
       // Do your fetching...
       await Future.delayed(const Duration(seconds: 2));
-      
+
       // ...and add the fetched nodes.
       item.children.addAll([
         TreeViewItem(
@@ -231,6 +231,7 @@ TreeView(
         TreeViewItem(
           content: const Text('Home Remodel'),
           value: 'home_remodel',
+          disabled: true,
           children: [
             TreeViewItem(
               content: const Text('Contractor Contact Info'),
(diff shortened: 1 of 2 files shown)


lib/src/controls/navigation/tree_view.dart
```diff
@@ -146,6 +146,10 @@ class TreeViewItem with Diagnosticable {
   /// item is going to be one of the selected items
   bool? selected;
 
+  /// Whether the current item is disabled.
+  ///
+  bool? disabled;
```
bdlukaa, line 151: I believe this can be non-nullable.

example/lib/screens/navigation/tree_view.dart
```diff
@@ -231,6 +231,7 @@ TreeView(
         TreeViewItem(
           content: const Text('Home Remodel'),
           value: 'home_remodel',
+          disabled: true,
```
bdlukaa, line 234: I think a new item can be created to showcase the disabled item functionality instead of using one that already exists.

bdlukaa (changes requested):
I would love to see what's the UI output of this as well. Thanks.

lib/src/controls/navigation/tree_view.dart
imxood: after add ```disabled: true```, the TreeViewItem will not be selected, and the son TreeViewItem can still be selected.

image

image

example/lib/screens/navigation/tree_view.dart
imxood, line 1: image

However, when the parent TreeViewItem is checked, all child TreeViewItems can still be selected.

In my application, if the parent TreeViewItem is disabled, then manually set the disabled property to true for all child elements.

bdlukaa: I believe this can be misleading. Maybe adding this to the documentation OR creating a new property (maybe `disableChildren`) that will disable the children if the parent is disabled?

PR #1068, opened 2024-05-31
feat: Add `Tab.color`, `Tab.selectedColor` and `Tab.outlineColor` to TabView

dmk-rib:
## Pre-launch Checklist

### DIFF
--- a/lib/src/controls/navigation/tab_view.dart
+++ b/lib/src/controls/navigation/tab_view.dart
@@ -92,8 +92,8 @@ class TabView extends StatefulWidget {
 
   /// The icon of the new button
   final IconData addIconData;
-  
-  /// The builder for the add icon. 
+
+  /// The builder for the add icon.
   ///
   /// This does not build the add button, only its icon.
   ///
@@ -741,6 +741,9 @@ class Tab with Diagnosticable {
     this.icon = const SizedBox.shrink(),
     required this.text,
     required this.body,
+    this.color,
+    this.selectedColor,
+    this.outlineColor,
     this.closeIcon = FluentIcons.chrome_close,
     this.onClosed,
     this.semanticLabel,
@@ -773,6 +776,15 @@ class Tab with Diagnosticable {
 
   /// The body of the view attached to this tab
   final Widget body;
+
+  /// The color of the tab.
+  final Color? color;
+
+  /// The color of the selected tab.
+  final Color? selectedColor;
+
+  /// The  outline color of the tab.
+  final Color? outlineColor;
 
   /// Whether the tab is disabled or not. If true, the tab will be greyed out
   final bool disabled;
@@ -928,9 +940,10 @@ class __TabState extends State<_Tab>
                   ),
             decoration: BoxDecoration(
               borderRadius: borderRadius,
-
               // if selected, the background is painted by _TabPainter
-              color: widget.selected ? null : backgroundColor,
+              color: widget.selected
+                  ? widget.tab.selectedColor
+                  : widget.tab.color ?? backgroundColor,
             ),
             child: () {
               final result = ClipRect(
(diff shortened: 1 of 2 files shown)


lib/src/controls/navigation/tab_view.dart
```diff
@@ -1062,6 +1076,15 @@ class _TabPainter extends CustomPainter {
...
       );
+
+    if (outlineColor != null) {
+      final outlinePaint = Paint()
+        ..color = outlineColor!
+        ..style = PaintingStyle.stroke
+        ..strokeWidth = 2.0;
+
+      canvas.drawPath(path, outlinePaint);
+    }
```
bdlukaa, line 1087: Can you show how this would look like? IIRC, this is not part of the official documentation or implementation.

lib/src/controls/navigation/tab_view.dart
```diff
@@ -774,6 +777,15 @@ class Tab with Diagnosticable {
...
   final Widget body;
 
+  /// The color of the tab.
+  final Color? color;
+
+  /// The color of the selected tab.
+  final Color? selectedColor;
+
+  /// The  outline color of the tab.
+  final Color? outlineColor;
```
bdlukaa, line 787: These, ideally, would have to be of type `ButtonState<Color>?`. The colors are not static and depend on the current state, specially on a desktop environment.

bdlukaa, line 787: The documentation for these is vague and incomplete, as well as the properties naming. What do you think of `backgroundColor` and `selectedBackgroundColor` instead of `color` and `selectedColor`, respectively?

lib/src/controls/navigation/tab_view.dart
```diff
@@ -928,9 +940,10 @@ class __TabState extends State<_Tab>
                   ),
             decoration: BoxDecoration(
               borderRadius: borderRadius,
-
               // if selected, the background is painted by _TabPainter
-              color: widget.selected ? null : backgroundColor,
+              color: widget.selected
+                  ? widget.tab.selectedColor
+                  : widget.tab.color ?? backgroundColor,
```
bdlukaa, line 946: The resolved `backgroundColor` must be applied even if `tab.selectedColor` is null.

```suggestion
              color: (widget.selected
                  ? widget.tab.selectedColor
                  : widget.tab.color) ?? backgroundColor,
```

bdlukaa (changes requested):

lib/src/controls/navigation/tab_view.dart
dmk-rib: dark
light

lib/src/controls/navigation/tab_view.dart
dmk-rib: I left these properties with the Color type because they are not part of the documentation, but can be very useful for specific user customizations. For example, provide a static tab color depending on the type of content.

lib/src/controls/navigation/tab_view.dart
```diff
@@ -774,6 +777,15 @@ class Tab with Diagnosticable {
...
   final Widget body;
 
+  /// The color of the tab.
+  final Color? backgroundColor;
+
+  /// The color of the selected tab.
+  final Color? selectedBackgroundColor;
+
+  /// The  outline color of the tab.
+  final Color? outlineColor;
```
bdlukaa, line 787:
```suggestion
  /// The background color of the tab.
  final Color? backgroundColor;

  /// The background color of the tab, if [selected] is `true`.
  final Color? selectedBackgroundColor;

  /// The outline color of the tab.
  final Color? outlineColor;
```

bdlukaa (approved):
LGTM!

PR #1080, opened 2024-07-02
fix: Update NumberBox formatting pattern for precision

### LINKED ISSUE #1079: 🐛 Numberbox format notwork when precision bigger then 4
**Describe the bug**
Use NumberBox to input 0.0001 and set precision, it will format display to 0

**To Reproduce**

```
NumberBox(
  precision:4,
  value:0.0001,
  onChanged:(){},
)
```

FourLeafTec:
fix #1079 .
By add NumberFormat pattern [L632](https://github.com/FourLeafTec/fluent_ui/blob/47fba0fde8a02e9d49546af48800c2926bdd24a5/lib/src/controls/form/number_box.dart#L632).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## next
+
+- fix: `NumberBox` format not work when precision bigger then 4.
+- feat: Add `NumberBox.pattern`, `NumberBox.formatter`, `NumberBox.format` to Numberbox.
+
 ## 4.9.0
 
 * fix: ¹ `DropDownButton.closeAfterClick` is now correctly applied. ² Added `MenuFlyoutItem.closeAfterClick`, which defaults to `true`. ([#1016](https://github.com/bdlukaa/fluent_ui/issues/1016))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## next
 
-- fix: `NumberBox` format not work when precision bigger then 4.
-- feat: Add `NumberBox.pattern`, `NumberBox.formatter`, `NumberBox.format` to Numberbox.
+- feat: ¹ `NumberBox` now supports precision greater than 4. ² Add `NumberBox.pattern`, `NumberBox.formatter`, `NumberBox.format` ([#1080](https://github.com/bdlukaa/fluent_ui/pull/1080))
 
 ## 4.9.0
 
--- a/lib/src/controls/form/number_box.dart
+++ b/lib/src/controls/form/number_box.dart
@@ -9,9 +9,9 @@ import 'package:math_expressions/math_expressions.dart';
 const kNumberBoxOverlayWidth = 60.0;
 const kNumberBoxOverlayHeight = 100.0;
 
-typedef FormatFunction = String? Function(num? number);
+typedef NumberBoxFormatFunction = String? Function(num? number);
 
-abstract interface class Formatter {
+abstract interface class NumberBoxFormatter {
   String format(dynamic number);
 }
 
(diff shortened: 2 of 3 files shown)


bdlukaa:
Great! Would you mind adding this to the CHANGELOG?

FourLeafTec:
Sure, but I just realized that this PR is a bit too simple, and I would like to add some more content:

1. Create properties to store the pattern instead of concatenating them each time
2. Allow users to pass in patterns to customize the format
3. Allow a format method to customize the formatting, but this will affect the parser, and I haven't figured this out yet. I would like to hear your thoughts.

It seems that the PR has also failed to merge, so I will find some time later to complete this part and then modify the CHANGLOG.. Is that okay?

bdlukaa:
Sounds okay to me!

bdlukaa:
@FourLeafTec let me known when this is ready for review!

lib/src/controls/form/number_box.dart
```diff
@@ -10,6 +9,12 @@ import 'package:math_expressions/math_expressions.dart';
 const kNumberBoxOverlayWidth = 60.0;
 const kNumberBoxOverlayHeight = 100.0;
 
+typedef FormatFunction = String? Function(num? number);
+
+abstract interface class Formatter {
+  String format(dynamic number);
+}
+
```
bdlukaa, line 17: These could be named `NumberBoxFormatCallback` and `NumberBoxFormatter`, respectivelly, to avoid confusion with other formatters

lib/src/controls/form/number_box.dart
```diff
@@ -95,8 +100,23 @@ class NumberBox<T extends num> extends StatefulWidget {
...
+  // The formatter for the number box. The formatter is used to format the
+  // number when the user input a value. If formatter set,precision, pattern and format
+  // must be null.
+  final Formatter? formatter;
+
+  /// The format function for the number box. The format function is used to
+  /// format the number when the user input a value. If format set,precision,
+  /// pattern and formatter must be null.
+  final FormatFunction? format;
```
bdlukaa, line 119:
```suggestion
  /// value. 
  ///
  /// If set, [pattern], [formatter] and [format] must be `null`. 
  ///
  /// Default is 2.
  final int? precision;

  /// The parttern for the number box. The pattern is used to format the number
  /// when the user inputs a value.
  ///
  /// If set, [precision], [formatter] and [format] must be `null`. 
  final String? pattern;

  /// The formatter for the number box. The formatter is used to format the
  /// number when the user inputs a value.
  ///
  /// If set, [pattern], [precision] and [format] must be `null`. 
  final Formatter? formatter;

  /// The format function for the number box. The format function is used to
  /// format the number when the user input a value.
  ///
  /// If set, [pattern], [formatter] and [precision] must be `null`. 
  final FormatFunction? format;
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## next
+
+- fix: `NumberBox` format not work when precision bigger then 4.
+- feat: Add `NumberBox.pattern`, `NumberBox.formatter`, `NumberBox.format` to Numberbox.
```
bdlukaa, line 4:
```suggestion
- feat: ¹ `NumberBox` now supports precision greater than 4. ² Add `NumberBox.pattern`, `NumberBox.formatter`, `NumberBox.format` ([#1080](https://github.com/bdlukaa/fluent_ui/pull/1080))
```

bdlukaa (changes requested):

lib/src/controls/form/number_box.dart
FourLeafTec: The `Formatter` should be named `NumberBoxFormatter`.

But the `FormatFunction` should be named `NumberBoxFormatFunction`.
This method is not a formatted callback, but rather it is used to format numbers.

My changes to the precision formatting function can be divided into three parts:

1. Simple specification of precision length: This allows users to directly specify the desired precision of a number without using any formatting methods.

2. Incorporating an inline formatting method: This method remains the same as before and is still the most commonly used way to format numbers.

3. A more complex and general Formatter class: This class provides more flexibility in formatting numbers, including the ability to customize the formatting rules and use different formatting styles.

Users can choose from any of these three approaches to achieve the desired formatting.

bdlukaa: Sure!

FourLeafTec:
@bdlukaa Review please.

bdlukaa (approved):

PR #1112, opened 2024-09-02
feat: [Picker] add `autoOpen` allow open panel auto when the widget is built

CorvusYe:
In some scenarios, we hope that time related forms will automatically pop up a panel when loaded to save the need for clicking again.

This PR affects `DatePicker`, `TimePicker`, and `Picker`

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - feat: `NavigationAppBar` `leading` widget is now a minimum of `kCompactNavigationPaneWidth` width instead of being fixed to this width ([#1103](https://github.com/bdlukaa/fluent_ui/pull/1103))
 - feat: Add `TabView.stripBuilder` ([#1106](https://github.com/bdlukaa/fluent_ui/issues/1106))
+- feat: [Picker] add `autoOpen` allow open panel auto when the widget is built. ([#1112](https://github.com/bdlukaa/fluent_ui/pull/1112))
 
 ## 4.9.1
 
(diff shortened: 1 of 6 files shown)


CorvusYe:
Thank you for your reply. 

I have some questions that I would like to discuss with you.
Does 'using keys' mean the following:

```dart
GlobalKey key = GlobalKey();
```

```dart
DatePicker(
  key: key,
  selected: v,
);
```

```dart
(key.currentState as dynamic).open();
```

If so, this calling method can indeed achieve the desired effect, but it seems more cumbersome. More encapsulation is needed to achieve parameter level logic control.

Moreover, `_DatePickerState` is private and can only be cast through "as dynamic" when we call the open method externally, it seems a bit strange.

I am not sure if my understanding is in line with popular thinking.

bdlukaa:
`autoOpen` can lead to unexpected behavior by the user. If the widget is ever entirely rebuilt, the panel would open as well. I prefer to leave this kind of control to the developer.

> Moreover, _DatePickerState is private and can only be cast through "as dynamic" when we call the open method externally, it seems a bit strange.

Ideally, we'd make the necessary changes to make the picker states public.

CorvusYe:
> `autoOpen` can lead to unexpected behavior by the user. If the widget is ever entirely rebuilt, the panel would open as well. I prefer to leave this kind of control to the developer.

You' re rignth. I overlooked this point. 

From this point of view, it would indeed be better for users to decide for themselves.

PR #1159, opened 2024-12-17
Fix visibility of Tab's close button when onClosed is null

diluculo:
Added missing `isClosable` property to `Tab` widget to match WinUI 3's [TabViewItem.IsClosable](https://learn.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.tabviewitem.isclosable?view=winui-2.8) implementation. 

The property defaults to `true`. When `isClosable` is `false`:
- Close button is not shown
- Keyboard shortcuts (Ctrl+W, Ctrl+F4) are disabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [next]
+
+- fix: hide Tab's close button when `onClosed` is null
+
 ## 4.10.0
 
 - fix: Add missing properties (`closeIconSize`, `closeButtonStyle`) in `debugFillProperties` and `InfoBarThemeData.merge` ([#1128](https://github.com/bdlukaa/fluent_ui/issues/1128)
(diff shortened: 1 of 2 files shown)


bdlukaa:
We use a different pattern in Flutter. If the `onClose` callback is null, it is automatically marked as disabled, therefore there is no need for another property.

diluculo:
Thank you for the feedback about `onClosed`.

I was actually trying to find a way to hide the close button in certain tabs. Looking at the current code, I noticed that the close button is shown even when `onClosed` is null.

I simplified the solution and just added a null check for `onClosed`.

image

In this image, the first tab has `onClosed = null` and with the added null check, its close button is now properly hidden while other tabs show their close buttons as expected.

bdlukaa (commented):
Can you add it to the changelog?

diluculo:
@bdlukaa, Done! I added the fix to the changelog and also squashed the commits into a single one.

bdlukaa (approved):

PR #1195, opened 2025-03-21
fix: tree view focus

### LINKED ISSUE #834: 🐛 TreeView's focus always starts at first or last item
**Describe the bug**
When using the arrow keys or the tab key to move the focus, the focus always starts on the first (down arrow, tab) or last item (up arrow) instead at the item that was invoked last:
Fluent_UI_Focus_with_arrows

**Expected behavior**
When using the keyboard to move the focus, the focus should start at the item that was invoked last, like can be seen in the WinUI 3 Gallery:
WinUI_Focus_with_arrows

black-06:
Partially fixes #834 
The focus should start at the tree item that was pressed last when using the keyboard

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 4.11.2
 
 - fix: Use correct scaffold background color when view is provided
+- fix: The focus should start at the tree item that was pressed last when using the keyboard
 
 ## 4.11.1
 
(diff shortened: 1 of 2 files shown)


bdlukaa:
Can you share a video of this working?

black-06:
20250322095308_rec_
​​​​
20250322095411_rec_

black-06:
I looked at the shortcut key behavior of WinUI 3 Gallery, `Tab` will directly switch to the next widget, `arrowUp/Down/Left/Right` will switch/expand item inside TreeView, but flutter `Tab` is similar to the `arrowDown`.
But at least now the focus will start with the most recently pressed item

Issue #1204, opened 2025-04-08
Add ability to customize `ListTile`'s built-in container margin


jacksonb-cs:
**Is your feature request related to a problem? Please describe.**
`ListTile`s have a built-in outer padding of `EdgeInsets.symmetric(horizontal: 4.0, vertical: 2.0)`. This makes it impossible to create list tiles that are directly adjacent to one another.

I see two potential solutions:

1. (Preferred) Add an `EdgeInsetsGeometry?` parameter to the widget's constructor with the current padding as the default value. This would add the ability to customize the margin without affecting existing projects.
2. Remove this default padding entirely. My problem with this solution is I don't exactly know how that would interact with the decorating container's surrounding `FocusBorder`.

PR #1205, opened 2025-04-08
Add `ListTile` margin property (#1204)

jacksonb-cs:
Added `ListTile.decorationMargin` parameter.
Added `kDefaultListTileDecorationMargin` constant.

### DIFF
--- a/lib/src/controls/surfaces/list_tile.dart
+++ b/lib/src/controls/surfaces/list_tile.dart
@@ -13,6 +13,11 @@ const kDefaultListTileShape = RoundedRectangleBorder(
   borderRadius: BorderRadius.all(Radius.circular(4.0)),
 );
 
+const kDefaultListTileDecorationMargin = EdgeInsetsDirectional.symmetric(
+  horizontal: 4.0,
+  vertical: 2.0,
+);
+
 enum ListTileSelectionMode {
   none,
   single,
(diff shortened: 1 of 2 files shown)


WinXaito (approved):
LGTM, maybe I would name the property `margin` instead of `decorationMargin`, but let @bdlukaa say what he think.

lib/src/controls/surfaces/list_tile.dart
```diff
@@ -168,6 +175,11 @@ class ListTile extends StatelessWidget {
   /// Defaults to [kDefaultListTilePadding]
   final EdgeInsetsGeometry contentPadding;
 
+  /// Padding applied between this tile's [FocusBorder] and outer decoration.
+  ///
+  /// Defaults to [kDefaultListTileMargin].
+  final EdgeInsetsGeometry? margin;
```
bdlukaa, line 181: Why is it nullable?

bdlukaa (changes requested):
Can you merge from master?

Also, add the margin parameter to debugFillProperties.

lib/src/controls/surfaces/list_tile.dart
jacksonb-cs: If no margin padding is desired, null is the most concise way to express that (as opposed to `EdgeInsets.all(0)`.

bdlukaa (approved):

PR #1218, opened 2025-05-15
Fix Undo entry in textbox context menu is doubled

WinXaito:
Because of this PR: #1212 Undo entry is doubled in menu (so it is in version 4.11.4)

Before:

image

After:

image

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [next]
+
+- fix: Undo entry was doubled in EditableText context menu ([#1218](https://github.com/bdlukaa/fluent_ui/pull/1218))
+
 ## 4.11.4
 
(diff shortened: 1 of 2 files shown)


bdlukaa:
I'm not sure this is the ideal solution. We need to find where the Undo action is being added twice to the context menu.

WinXaito:
Looks it's because of that:

(line 131)
```dart
.followedBy(buttonItems.where((item) =>
            item.type == ContextMenuButtonType.custom && item.label == 'Undo'))
```

and right after (line 135):
```dart
.followedBy(buttonItems
            .where((item) => item.type == ContextMenuButtonType.custom))
```

So there is the duplication.
That's why I add `item.label != 'Undo'`, juste because it was already using `item.label == 'Undo'` before.

Issue #236, opened 2022-03-23
Calendars Rework


bdlukaa:
The current implementation of both date and time pickers are old and have many issues.

- [x] They still use `Acrylic` as their base implementation, making it look like it's not native (https://github.com/bdlukaa/fluent_ui/pull/241)
- [x] They only saves last changed dialer (https://github.com/bdlukaa/fluent_ui/issues/161)
- [x] They do not make use of the `FluentLocalizations` (https://github.com/bdlukaa/fluent_ui/issues/213)
- [x] The popup isn't positioned correctly over the button when there is space. Instead, it's either above or below.

---

- [x] Implement `CalendarView` (https://github.com/bdlukaa/fluent_ui/pull/1232)
- [ ] Implement `CalendarDatePicker`, based on `CalendarView`

henry2man:
Following DRY principles, can we ever consider use something curated like [table_calendar](https://github.com/aleksanderwozniak/table_calendar) as a base?

I have some experience with this one and this could simplify the ulterior development if we are able to render the calendar as expected in the spec.

@bdlukaa Do we have any specific policy about third party packages?

bdlukaa:
As long as it's able to allow us to customizate it for our needs, I don't see any problem!

henry2man:
@bdlukaa So, in order to make things clear prior to any development:

- We want to create a new `CalendarView`, potentially based on [table_calendar](https://github.com/aleksanderwozniak/table_calendar). 
- We want to **substitute** current calendar picker with one that shows a date and pops a flyout with a CalendarView inside.
- For this very first implementation we're going to have a minimal API:  
  - `CalendarView`
    - `initialValue`: Show a defined month & Focus into that day (or `DateTime.now()` if not set)
    - mandatory `onSelect` callback: function to be called when a day is selected
    - `weekStartDay`: First day of week (normally Sunday or Monday, depending on localization), defaults to Sunday.
    - UI: Change month using "up" and "down" buttons
    - UI: Use current locale to show current month & weekday tiles and also for formatting.
  - `CalendarDatePicker`
    - Display a form field with `initialValue` date (or `DateTime.now()` if not set)
    - On click, display a flyout with a `CalendarView` inside
    - When a day is clicked, hide the flyout and execute mandatory `onChange` callback

henry2man:
@bdlukaa  Can we close this issue? What is missing?

bdlukaa:
`CalendarView`

[...]

PR #1232, opened 2025-06-17
Add WinUI3 CalendarView widget

Asmitta-01:
This PR is adding the [CalendarView](https://learn.microsoft.com/en-us/windows/apps/design/controls/calendar-view) widget to the package.

## Pre-launch Checklist

The widget works like the one in WinUI demo software but i noticed some points that might be seen:
- Add an animation when switching [display modes](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.calendarview.displaymode)
- Make the Decade and year display mode to be scrollable

I want to know if there're other issues

### DIFF
--- a/lib/src/controls/pickers/calendar_view.dart
+++ b/lib/src/controls/pickers/calendar_view.dart
@@ -362,6 +362,7 @@ class CalendarViewState extends State<CalendarView> {
   void didUpdateWidget(covariant CalendarView oldWidget) {
     super.didUpdateWidget(oldWidget);
 
+    // Update the min/max display years if the date range changes
     if (widget.displayDateStart != oldWidget.displayDateStart ||
         widget.displayDateEnd != oldWidget.displayDateEnd) {
       final currentDate = DateTime.now();
@@ -371,8 +372,19 @@ class CalendarViewState extends State<CalendarView> {
       _maxDisplayedYear =
           widget.displayDateEnd?.year ??
           currentDate.year + (_defaultDisplayedYearCount ~/ 2);
+
+      // Clamp anchorMonth to new range if needed
+      final minDate = widget.displayDateStart ?? DateTime(_minDisplayedYear);
+      final maxDate =
+          widget.displayDateEnd ?? DateTime(_maxDisplayedYear, 12, 31);
+      if (_anchorMonth.isBefore(minDate)) {
+        _anchorMonth = minDate;
+      } else if (_anchorMonth.isAfter(maxDate)) {
+        _anchorMonth = DateTime(maxDate.year, maxDate.month, 1);
+      }
     }
 
+    // Handle selection mode changes by resetting selection state
     if (widget.selectionMode != oldWidget.selectionMode) {
       _selectedStart = widget.initialStart;
       _selectedEnd = widget.selectionMode == CalendarViewSelectionMode.range
@@ -388,6 +400,7 @@ class CalendarViewState extends State<CalendarView> {
       return;
     }
 
+    // Update selection when initial dates change
     if (widget.initialStart != oldWidget.initialStart ||
         (widget.selectionMode == CalendarViewSelectionMode.range &&
             widget.initialEnd != oldWidget.initialEnd)) {
@@ -404,31 +417,26 @@ class CalendarViewState extends State<CalendarView> {
       _notifySelectionChanged();
     }
 
+    // Validate and update selections when min/max dates change
     if (widget.minDate != oldWidget.minDate ||
         widget.maxDate != oldWidget.maxDate) {
-      bool selectionWasInvalidated = false;
+      final wasStartInvalid =
+          _selectedStart != null && !_isDateWithinBounds(_selectedStart!);
+      final wasEndInvalid =
+          _selectedEnd != null && !_isDateWithinBounds(_selectedEnd!);
+      final originalCount = _selectedMultiple.length;
+      _selectedMultiple.removeWhere((date) => !_isDateWithinBounds(date));
+      final wasMultipleInvalid = originalCount != _selectedMultiple.length;
 
-      if (_selectedStart != null && !_isDateWithinBounds(_selectedStart!)) {
-        _selectedStart = null;
-        selectionWasInvalidated = true;
-      }
-      if (_selectedEnd != null && !_isDateWithinBounds(_selectedEnd!)) {
-        _selectedEnd = null;
-        selectionWasInvalidated = true;
-      }
-      if (_selectedMultiple.isNotEmpty) {
-        final originalCount = _selectedMultiple.length;
-        _selectedMultiple.removeWhere((date) => !_isDateWithinBounds(date));
-        if (_selectedMultiple.length != originalCount) {
-          selectionWasInvalidated = true;
-        }
-      }
+      if (wasStartInvalid) _selectedStart = null;
+      if (wasEndInvalid) _selectedEnd = null;
 
-      if (selectionWasInvalidated) {
+      if (wasStartInvalid || wasEndInvalid || wasMultipleInvalid) {
         _notifySelectionChanged();
       }
     }
 
+    // Update display mode if it changed and differs from current
     if (widget.initialDisplayMode != oldWidget.initialDisplayMode &&
         widget.initialDisplayMode != _displayMode) {
       _displayMode = widget.initialDisplayMode;


bdlukaa (changes requested):
Thank you so much for this PR! This fixes part of #236.

My main feedback is scrolling. I see `PageView` was used to make scrolling possible, but it doesn't match or mimic the native counterpart behavior.

Current behavior:


https://github.com/user-attachments/assets/956e5595-65ae-4c89-adbf-8d2225877253


Expected behavior:


https://github.com/user-attachments/assets/cae23b9f-559d-4be4-887d-5d06a72a9685



Additionally, here are the resources for this control:
* Spec: https://github.com/microsoft/microsoft-ui-xaml/blob/main/specs/CalendarView/CalendarViewSpec1.md 
* Resources: https://github.com/microsoft/microsoft-ui-xaml/blob/fb7a83b668baa612b5cc594678746dd8c1f8d8bd/src/controls/dev/CommonStyles/CalendarView_themeresources.xaml

We will want to use the same resources as native. You can access all the color resources using `FluentTheme.of(context).resources.[resource name]`. The resources also provide sizing, padding and margins of the control. The goal is to have it as close as native as possible.

Other things to notice:

* The size of the inner buttons doesn't match native, specially for the year and decade views;
* The margin between the buttons doesn't match native;
* The header title and buttons sizes don't match native;

Overall, I like the PR.

Asmitta-01:
> My main feedback is scrolling. I see PageView was used to make scrolling possible, but it doesn't match or mimic the native counterpart behavior.

I had an issue with PageView so instead i built a full list for the calendar(now in a ListView) and i use the offset to determine the currenth month. But i still have issues: The ListView seems heavier (i want to display 1000 weeks) and the listener on it works _sometimes_. 
For example 1000 weeks = 19 years(19.23) but my list have no limit (i even reached 1680 one time). I know the main issue may be might code but i want to leave ListView now.

I want to go back PageView. Setting `pageSnapping` to false makes the scrolling feel like native but  i have duplicated rows.  Any suggestion to improve it  ? 😓

bdlukaa:
I think the solution to the scrolling behavior is making an infinite scroll list in both dimensions. It is possible with Flutter using a CustomScrollView: https://github.com/flutter/flutter/issues/20608. I am experimenting with this.

bdlukaa:
I changed the scrolling widget from `PageView` to a `CustomScrollView`. With the way a calendar work, with dates overlapping months, a page view isn't the ideal solution here. Instead of a page view, I used two grid views, creating an infinite scrolling view for both directions: up and down (https://github.com/flutter/flutter/issues/20608), which creates a negative and a positive index for the children, which allow us to calculate the date based on the index and the anchor date. In the end we got a much closer-to-native experience with enhanced performance.

Additionally, I exposed `CalendarViewState`, which exposes the navigation methods:

```dart
final calendarKey = GlobalKey<CalendarView>();

CalendarView(
  key: calendarKey,
)

calendarKey.currentState?.firstDayOfWeek; // Gets the first day of the week
calendarKey.currentState?.visibleDate;  // Get the current visible date, whether by scrolling or navigating
calendarKey.currentState?.navigateToMonth(date); // Navigates to a specific month. 
calendarKey.currentState?.stepMonth(offset: 1);  // Steps a month by offset
calendarKey.currentState?.navigateToYear(date); // Navigates to a specific year
calendarKey.currentState?.stepYear(offset: 1); // Steps a year by offset
```

We may consider adding a controller or a listener that allows listening to the current visible date, which is constantly updated by scrolling.

From this point, only the decades view needs a rework.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1084 @@
...
+    if (_selectedStart == null || _selectedEnd == null) return false;
+    return !day.isBefore(_selectedStart!) && !day.isAfter(_selectedEnd!);
+  }
+
+  void _onDayTapped(DateTime day, bool inScope) {
+    if (widget.selectionMode == CalendarViewSelectionMode.none) return;
+    if (!inScope && !widget.isOutOfScopeEnabled) return;
+
+    setState(() {
+      if (widget.selectionMode == CalendarViewSelectionMode.single) {
```
bdlukaa, line 341: I think a switch statement would be better here.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1084 @@
...
+            endDate: _selectedMultiple.length > 1
+                ? _selectedMultiple.last
+                : null,
+          ),
+        );
+      }
+    });
+  }
+
+  int get firstDayOfWeek {
```
bdlukaa, line 395: Documentation missing.
```suggestion
  /// Gets the first day of the week based on the current locale.
  int get firstDayOfWeek {
```

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1084 @@
...
+      mainAxisSize: MainAxisSize.min,
+      spacing: 4,
+      children: [
+        _buildHeader(),
+        const Divider(
+          style: DividerThemeData(horizontalMargin: EdgeInsets.zero),
+        ),
+        const SizedBox(height: 4),
+        SizedBox(
+          height: 4 * 68.0,
```
bdlukaa, line 615: Maybe add a `widget.yearsPerView`, just like the month view.

Asmitta-01:
In this [image](https://learn.microsoft.com/en-us/windows/apps/design/controls/images/calendar-view-month-elements.png) blackout dates seem disabled, like out-of-scope dates but in the [specs](https://github.com/microsoft/microsoft-ui-xaml/blob/main/specs/CalendarView/CalendarViewSpec1.md#blacked-out-styling) they're strikethrough. I don't know which one we should follow.

-------------------------------------------------

I talk about it because i think you're working on the scolling issue(don't know if we should on it separately).

bdlukaa:
@Asmitta-01 you should follow the latest specs. The image you provided is from Windows 10.

example/lib/screens/forms/calendar_view.dart
Asmitta-01, line 1: @bdlukaa I had doubts about the good path for `LanguageLocale` class so i left it here. Hope you'll find the best location.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+  /// Any date respecting this rule will be considered as a "blackout" date, meaning
+  /// users will not be able to select it. This is useful for preventing selection
+  /// of holidays, weekends, or other restricted dates.
+  ///
+  /// Example:
+  /// ```dart
+  /// // To blackout all weekends
+  /// blackoutRule: (date) => date.weekday == DateTime.saturday || date.weekday == DateTime.sunday,
+  /// ```
+  final bool Function(DateTime date)? blackoutRule;
```
bdlukaa, line 141: This could be a typedef.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+  final DateTime? minDate;
+
+  /// The maximum date that can be selected.
+  final DateTime? maxDate;
+
+  /// The last date in the date range that is available in the calendar..
+  final DateTime? displayDateEnd;
+
+  /// The first date that is available in the calendar.
+  final DateTime? displayDateStart;
```
bdlukaa, line 107: Documentation could be improved saying the difference between these properties. We could also create a separate object called `DateBounds`(?) and reduce these properties.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+  final _monthScrollController = ScrollController();
+  final _yearScrollController = ScrollController();
+  final _decadeScrollController = ScrollController();
+
+  final _scrollBehavior = const ScrollBehavior().copyWith(
+    scrollbars: false,
+    physics: const BouncingScrollPhysics(
+      decelerationRate: ScrollDecelerationRate.fast,
+    ),
+  );
```
bdlukaa, line 273: This could be a customizable property in the widget.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+  @override
+  void dispose() {
+    _monthScrollController.dispose();
+    _yearScrollController.dispose();
+    _decadeScrollController.dispose();
+    super.dispose();
+  }
+
+  @override
+  void didUpdateWidget(covariant CalendarView oldWidget) {
```
bdlukaa, line 346: This entire function needs to be property reviewed and documented.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+  }
+
+  void _monthScrollListener() {
+    final pixels = _monthScrollController.position.pixels;
+    final rowIndex = (pixels / _rowHeight).round();
+    if (_displayMode == CalendarViewDisplayMode.month) {
+      final newScrollDate = _daysFromAnchor((rowIndex * 7 + 10).toInt());
+      if (_shouldUpdateDate(newScrollDate)) {
+        // todo(kv): work pretty well when adding 10 days.
+        //  It is maybe due because of bug of the duplicated week in June.
```
bdlukaa, line 484: Explore this TODO.

lib/src/controls/pickers/calendar_view.dart
```diff
@@ -0,0 +1,1565 @@
...
+        break;
+      case CalendarViewDisplayMode.year:
+        label = date.year.toString();
+        break;
+      case CalendarViewDisplayMode.decade:
+        final startYear = (date.year ~/ 10) * 10;
+        final endYear = startYear + 9;
+        label = '$startYear - $endYear';
+        break;
+    }
```
bdlukaa, line 1234: could be an inline switch

bdlukaa (commented):
Looks good overall. I'd just make improvements to the code to make it cleaner and more readable. It may be worth considering moving parts to a different file.

lib/src/controls/pickers/calendar_view.dart
Asmitta-01: I don't understand...

bdlukaa (approved):

PR #1328, opened 2026-05-26
feat: add `ListTile.tileGradient`

jacksonb-cs:
Add the ability to provide a gradient to `ListTile`.

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [next]
+
+- fix: `Tooltip` now correctly uses constructor-given style
+
 ## 4.15.1
 
 - fix: `NavigationView` auto display mode no longer shows a brief overlay when resizing from minimal to compact mode ([#1316](https://github.com/bdlukaa/fluent_ui/pull/1316))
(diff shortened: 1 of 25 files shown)


bdlukaa:
Instead of a gradient, what about allowing a `Decoration` object? Like `Container.color` and `Container.decoration`. It'll give more flexibility and avoid adding a new parameter for each decoration parameter needed.

jacksonb-cs:
Since I'll remove `ListTile.shape` in order to allow passing any type of `Decoration` (defaulting to the current `ShapeDecoration` and default list tile shape if no decoration is provided), this will be a breaking change. Should I somehow specifically indicate that in the changelog?

jacksonb-cs:
Removing `ListTile.shape` will also require removing its `DiagnosticsNode` from `debugFillProperties`. I would add the decoration in its stead, but the default decoration has to take the color resolved from theme data, so I cannot give a proper `FlagProperty.defaultValue`.

Since `tileColor`, which is resolved from the theme, is not included in `debugFillProperties`, I would assume I can just exclude the decoration as well; however, I'm unfamiliar with these debug properties, so let me know if I should do otherwise.

PR #1340, opened 2026-07-05
Allow TitleBar to not respond to onDoubleTap to fix visible delay on clicks

artdedeco:
The new `TitleBar` control, which replaces the old `NavigationAppBar`, currently uses a `GestureDetector` to respond to pan (drag) and double tap/click events. However, `GestureDetector` has a [known, currently unresolved, issue](https://github.com/flutter/flutter/issues/110300) where it adds a 300ms delay to single tap events when you listen to `onDoubleTap`, **thus rendering `TitleBar` unusable** in any desktop, mouse & keyboard environment.

There are a few different avenues for fixing this on the library side, notably: 1) depend on a lower level gesture detection abstraction and choose our own algo for differentiating between single and double clicks  - perhaps one better suited for mouse, given the target audience of the library; or 2) redesign the control so that gesture detection does not apply to the whole thing (which would have the added benefit of better aligning to Windows UX standards -- drag should not work when clicking on app icon or caption buttons, for example).

Given that both of those options would not only require non-trivial amount of work, but also a discussion phase, I'm proposing this change as a quick mitigation, so that at the very least it unblocks teams currently blocked by this, such as mine. The change is a one liner: from `onDoubleTap: () => onDoubleTap?.call()` to `onDoubleTap: onDoubleTap`; the idea being that if the caller provides a `null` func to `onDoubleTap`, the `onDoubleTap` callback from `GestureDetector` is never registered.

I categorized the commit as `refactor(perf)` following precedence on the changelog file, and given that the change does not alter the contract nor does it cause any externally visible change in behavior for the caller (beside the improved perf).

## Pre-launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
   `MenuBar` now uses `RawMenuBar` under the hood.
 - feat: `Tooltip.enableTapToDismiss` and `Tooltip.onTriggered` ([#1338](https://github.com/bdlukaa/fluent_ui/pull/1338))
   `Tooltip` now uses `RawTooltip` under the hood.
+- refactor(perf): `TitleBar` now doesn't listen for `onDoubleTap` if the caller doesn't provide a callback. Listening for `onDoubleTap` causes a 300ms delay on single taps (see flutter/flutter#110300)
 
 ## 4.16.0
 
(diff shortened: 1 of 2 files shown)


artdedeco:
I just noticed #1298 was responsible for adding this (support for double click callback) - I had thought it was a more fundamental design decision to support it from the start; Instead, it seems more like it was just an oversight in that implementation. I guess that makes this fix I'm proposing acceptable even long-term.

bdlukaa (approved):

